Issue #10, opened 2022-12-24
[Windows] Black flicker on some GPUs/devices when using H/W rendering


alexmercerind:
Feedback needed.

alexmercerind:
Personally speaking, I have entry-level AMD Ryzen 3 2200U processor with integrated Raden Vega 3 Mobile Graphics.

I don't see any flicker at all. Only 4K 60 FPS videos seem to be flickering, but even 4K 30 FPS are playing fine.
Please comment your experience.

alexmercerind:
No objections so far. Closing.
No problems on other GPUs/devices I tested either.

Quackdoc:
I was able to replicate it on my rx580, it seems like it might be random 
[info.txt](https://github.com/alexmercerind/media_kit/files/10302894/info.txt)

alexmercerind:
> I was able to replicate it on my rx580, it seems like it might be random
> [info.txt](https://github.com/alexmercerind/media_kit/files/10302894/info.txt)

Thanks!
- You attempted to play the videos present on the right-side, right?
- From the logging file, are you running on higher than 60 FPS?

Quackdoc:
Im running 60fps on my monitor, 

I did play the videos on the right side of the screen and they rarely caused flickering to occur. however downloading a longer video from youtube and loading that up did consistently cause the flickering to occur, I think the higher the resolution the more it occurs. 

I did try disabling hwdec, but that didn't seem to have any bearing on the flickering at all. 

I did also add file-log to the mpv string, and the logs it pumped out didn't seem to be correlate to the flashing at all.

I did try to test on linux, but the same program you have doesn't seem to compile, so I replaced vlc_dart with mediakit in a private application i have accsess to, but sadly it doesn't seem like embedding code is wired up, though it does create an mpv context just fine at least.

alexmercerind:
Thanks!

> I did play the videos on the right side of the screen and they rarely caused flickering to occur. however downloading a longer video from youtube and loading that up did consistently cause the flickering to occur, I think the higher the resolution the more it occurs.

On my device, only 4K 60 FPS videos flicker black.
I don't have much idea either. However, it is really important to resolve this issue.

> I did try to test on linux, but the same program you have doesn't seem to compile, so I replaced dart_vlc with media_kit in a private application I have access to, but sadly it doesn't seem like embedding code is wired up, though it does create an mpv context just fine at least.

Yeah, Linux implementation is not done yet.
Currently focusing on Windows.

[...]

PR #11, opened 2022-12-30
Flickerfix mostly

Quackdoc:
so turns out the brute force way can be the best way, the fixes 99% of flickers, to the degree that loading an 8k video with an gpu stress test running in the foreground is, while still not watchable, not flickering at all... mostly 

every now and then there is a rare flicker, however im fairly convinced that this is because of skia itself, as other elements of the app become troublesome anyways, like animations freezing. and I really don't think there is anything we can do with that.

this MIGHT be troublesome due to the extra glfinishes. meaning high refreshrate content might suffer from this, however I don't have a display to test that with unfortunately.

also fixes resize flicker that I was experiencing. 

all in all, 4k60 content is now flicker free

hopefully Addresses #10

### DIFF
--- a/media_kit_core_video/windows/angle_surface_manager.cc
+++ b/media_kit_core_video/windows/angle_surface_manager.cc
@@ -61,7 +61,7 @@ void ANGLESurfaceManager::MakeCurrent(bool value) {
     eglMakeCurrent(display_, surface_, surface_, context_);
   } else {
     glFinish();
-    eglMakeCurrent(display_, surface_, surface_, context_);
+    eglMakeCurrent(display_, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT);
   }
 }
 
--- a/media_kit_core_video/windows/angle_surface_manager.cc
+++ b/media_kit_core_video/windows/angle_surface_manager.cc
@@ -61,7 +61,7 @@ void ANGLESurfaceManager::MakeCurrent(bool value) {
     eglMakeCurrent(display_, surface_, surface_, context_);
   } else {
     glFinish();
-    eglMakeCurrent(display_, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT);
+    eglMakeCurrent(EGL_NO_DISPLAY, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT);
   }
 }
 
--- a/media_kit_core_video/windows/angle_surface_manager.cc
+++ b/media_kit_core_video/windows/angle_surface_manager.cc
@@ -43,7 +43,6 @@ HANDLE ANGLESurfaceManager::HandleResize(int32_t width, int32_t height) {
   }
   width_ = width;
   height_ = height;
-  glFinish();
   // Create new Direct3D texture & |surface_| preserving previously created
   // |display_| & |context_| from the constructor.
   Initialize();
@@ -57,11 +56,16 @@ void ANGLESurfaceManager::SwapBuffers() {
 
 void ANGLESurfaceManager::MakeCurrent(bool value) {
   if (value) {
+#ifdef ENABLE_GL_FINISH_SAFEGUARD
     glFinish();
+#endif
     eglMakeCurrent(display_, surface_, surface_, context_);
   } else {
+#ifdef ENABLE_GL_FINISH_SAFEGUARD
     glFinish();
-    eglMakeCurrent(EGL_NO_DISPLAY, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT);
+#endif
+    eglMakeCurrent(EGL_NO_DISPLAY, EGL_NO_SURFACE, EGL_NO_SURFACE,
+                   EGL_NO_CONTEXT);
   }
 }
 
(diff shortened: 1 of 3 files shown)


Quackdoc:
@alexmercerind so it turns out, the `fix woops` commit I made, makes the issues appear again, and I think I figured it kinda out????

when running the demo app

```cpp
eglMakeCurrent(display_, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT);  // bad flicker
eglMakeCurrent(EGL_NO_DISPLAY, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT); // rare flicker
eglMakeCurrent(display_, surface_, surface_, context_); // rare flicker
```

however when I run one of the apps im working on 

```cpp
eglMakeCurrent(display_, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT);  // bad flicker
eglMakeCurrent(EGL_NO_DISPLAY, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT); // medium flicker
eglMakeCurrent(display_, surface_, surface_, context_); // medium flicker
```

alexmercerind (commented):
Hi!

Thanks for the fixes. As far as I can think, I believe this is the best intended approach we can have for syncing.

I was thinking to use `glFenceSync` & family for syncing, but it seems it's [not available in OpenGL ES 2.0 standard](https://stackoverflow.com/a/12157120/12825435) anyway.
So, `glFinish` is the only way to go.

<hr>

BTW, is [this `glFinish`](https://github.com/alexmercerind/media_kit/blob/083f128876f4a156a61c70a73c93558b808e3c1f/media_kit_core_video/windows/video_output.cc#L157) still required? I suppose no.

... Because it's called "twice" (once outside & once inside `ANGLESurfaceManager::MakeCurrent`) right before leaving the context.

media_kit_core_video/windows/angle_surface_manager.cc
```diff
@@ -43,6 +43,7 @@ HANDLE ANGLESurfaceManager::HandleResize(int32_t width, int32_t height) {
   }
   width_ = width;
   height_ = height;
+  glFinish();
```
alexmercerind, line 46: If you don't mind, can I review this out?

How about we keep all `glFinish` calls behind a macro in `angle_surface_manager.cc` like

```c

...

#define SAFEGUARD_GLFINISH

...

#ifdef SAFEGUARD_GLFINISH
  glFinish();
#endif

```

media_kit_core_video/windows/angle_surface_manager.cc
```diff
@@ -56,9 +57,11 @@ void ANGLESurfaceManager::SwapBuffers() {
 
 void ANGLESurfaceManager::MakeCurrent(bool value) {
```
alexmercerind, line 58: Now that releasing context is not required (based on your last comment), let's remove the argument from this method.

And only use it to acquire context.

```suggestion
void ANGLESurfaceManager::MakeCurrent() {
```

Thus, remove entirety of this (if it goes well):

https://github.com/alexmercerind/media_kit/blob/083f128876f4a156a61c70a73c93558b808e3c1f/media_kit_core_video/windows/video_output.cc#L147-L158

Quackdoc:
it shouldn't but it *feels* like it helps. I would need to sit down and go over it more. strictly speaking thought it really shouldn't be doing anything. but after about 10 minutes or so of watching looping animations it *seems* like it helps

alexmercerind (changes requested):
My best gest goes that `glFinish` is needed before acquiring context & rendering (using libmpv).

Instead of after drawing frame & then calling `glFinish`, like I did.

media_kit_core_video/windows/angle_surface_manager.cc
Quackdoc: sounds good to me, this one I don't even think is necessary, I only added it since it shouldn't be called that much and iirc remeber reading that someone had flicker issue on resize

media_kit_core_video/windows/angle_surface_manager.cc
Quackdoc: Ill test it for sure when I can

Quackdoc: after half an hour of the two side by side, it's best to keep it as is, it's quite a bit more stable? it flickers left often.

media_kit_core_video/windows/angle_surface_manager.cc
Quackdoc: yeah I don't see much of a benefit to keeping this one, all it does is make it feel less sluggish, I couldn't observe any flickering on resize.

alexmercerind:
Now why is this crashing on me. I'mma cry now.

alexmercerind (approved):
LGTM!

I added ENABLE_GL_FINISH_SAFEGUARD macro for all these glFinish calls for now. Now that there are no black flickers anymore & everything is working fine, I can move onto next platform's implementation _i.e._ Linux.

Also:

I will expose a new parameter within Dart to [`VideoController.create` method](https://github.com/alexmercerind/media_kit/blob/083f128876f4a156a61c70a73c93558b808e3c1f/media_kit_core_video/lib/video_controller.dart#L72-L77) for disabling H/W rendering & enforcing S/W renderer. Just in case one wants to use it.

<hr>

_I did a_ flutter clean _& it's working fine. I mixed build files between two channels likely._

alexmercerind:
I also sent you invitation for this repository, which you hopefully accepted.
Please make sure to follow: www.conventionalcommits.org specification in your commit messages.

Thanks!

PR #16, opened 2023-01-07
feat: Add URI to test application

Quackdoc:
Add a test mode where the user can paste URIs directly. 

I was pretty conflicted on whether or not to add the the ability to load audio files directly via setting properties.
but in general I think developers being able to set the properties is a very useful feature to have. and is a good idea to somewhat showcase it. 

as for the act of loading dedicated audio files with videos, this in general is a very useful feature to have.

### DIFF
--- a/media_kit_test/lib/main.dart
+++ b/media_kit_test/lib/main.dart
@@ -58,6 +58,21 @@ class HomeScreen extends StatelessWidget {
           ),
           ListTile(
             title: const Text(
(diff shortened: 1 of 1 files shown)


alexmercerind (approved):
SGTM!

Feel free to merge. I'm actually busy with my semester examinations till the end of this month. I will make further advancements & implementation(s) for other platforms after returning.

Thanks.

Quackdoc:
Best of luck with exams!

PR #19, opened 2023-01-14
fix(win): resize texture & EGLSurface on render thread

### LINKED ISSUE #17: [Windows] Simultaneous Video & Player crash on texture/surface resize
The app crashes when I try to use 5 videos simultaneously. 

I'm getting the following error:

alexmercerind:
Closes #17.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
           submodules: "true"
       - uses: subosito/flutter-action@v1
         with:
-          channel: "stable"
+          channel: "master"
       - run: flutter pub get
       - run: flutter build windows --verbose
       - uses: actions/upload-artifact@v1
--- a/media_kit_core_video/lib/video.dart
+++ b/media_kit_core_video/lib/video.dart
@@ -117,9 +117,29 @@ class _VideoState extends State<Video> {
                           return SizedBox(
                             width: rect.width,
                             height: rect.height,
-                            child: Texture(
-                              textureId: id,
-                              filterQuality: widget.filterQuality,
+                            child: Stack(
+                              children: [
+                                const SizedBox(),
+                                Positioned.fill(
+                                  child: Texture(
+                                    textureId: id,
+                                    filterQuality: widget.filterQuality,
+                                  ),
+                                ),
+                                // Keep the |Texture| hidden before the first frame renders. In native implementation, if
+                                // no default frame size is passed (through VideoController), a starting 1 pixel sized
+                                // texture/surface is created to initialize the render context & check for H/W support.
+                                // This is then resized based on the video dimensions & accordingly texture ID, texture,
+                                // EGLDisplay, EGLSurface etc. (depending upon platform) are also changed.
+                                // Just don't show that 1 pixel texture to the UI.
+                                // NOTE: Unmounting |Texture| causes the |MarkTextureFrameAvailable| to not do anything.
+                                if (rect.width == 1.0 && rect.height == 1.0)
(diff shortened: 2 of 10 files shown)


alexmercerind:
Seems I used an API which isn't present on stable channel yet. Let's see what can I do.

Quackdoc:
The fix as of [17418a1](https://github.com/alexmercerind/media_kit/pull/19/commits/17418a13a11de3e3c91128c599da156cc9dddff3) works on my end quite well, I dont have any instability using glfinish, and rapidly swapping between videos on the test application no longer crashes either.

alexmercerind:
> The fix as of [17418a1](https://github.com/alexmercerind/media_kit/pull/19/commits/17418a13a11de3e3c91128c599da156cc9dddff3) works on my end quite well, I dont have any instability using glfinish, and rapidly swapping between videos on the test application no longer crashes either.

Thanks for the confirmation!
I'm just making sure that things keep working with stable channel as-well.

Merging in a while.

alexmercerind:
This commit message explains the fix: [`237794a`](https://github.com/alexmercerind/media_kit/pull/19/commits/237794a485cc09191c319b78609c90147266f24c).

Yesterday I got confused by fallback S/W rendering. 😄
I thought "this isn't crashing anymore" (because it was using S/W rendering) & pushed. I was in hurry, left it & couldn't fix later.

Quackdoc:
it does at least work on beta channel, so if needed the PR can sit till beta becomes stable. certainly wouldn't be ideal, for sure however...

alexmercerind:
I feel this important to merge because:
- The package isn't released on pub.dev either-way & everyone knows that I'm still working on it.
  So, requiring `master` channel shouldn't be that surprising.
- Being able to play multiple videos simultaneously is important. Video playback is taken for granted by everyone.
  Reaching that level of stability & performance is the goal.
  
I just made a workaround to avoid using that new API. Now build succeeds on stable channel but I'm getting a crash. Let's see.

Quackdoc:
I agree, considering it's WIP status, I don't think it's an issue at all, especially since a work around exists

alexmercerind:
First of all, this fix works **very-very well & stable on master or beta channels  (Flutter 3.7.x)**.

There are no flickers & no crashes with either solo or simultaneous video playback. It doesn't matter how hard I "bully" the test application. Resize in `EGLSurface` & creation of new texture due to change in video have no effect either. There are no race conditions in my code now. I've set mutex(s) everywhere. Now, resizing is done on rendering thread itself (before rendering the newly sized frame).

Download Artifact: https://github.com/alexmercerind/media_kit/actions/runs/3919351537 [For Testing]

<hr>

**I decided to drop the support for stable channel  (Flutter 3.3.x)** & remove that hack.

Nothing seems to work in a stable manner. If we decide to keep support for Flutter 3.3, an option can be to hardcode the surface/texture size e.g. to 1080p or something. However, there is no official way to "query" Flutter version in-code. But, with current H/W rendering implementation (where texture size changes based on video) Flutter 3.3 just randomly crashes sometimes when unregistering texture, sometimes when registering new texture, sometimes when rendering. Sometimes crash can be seen when going back i.e. disposing the `VideoController`.

These originate within `flutter_windows.dll`. It could be an incorrect usage by me, but now that there are no races within my code, I don't think there's anything that can be done for Flutter 3.3 (when newer Flutter versions are working great).

Maybe these have made difference in Flutter 3.7 ? :
I just hope they don't "touch it much" now that it works fine.

https://github.com/flutter/engine/pull/25349
https://github.com/flutter/engine/pull/33688

PR #27, opened 2023-01-18
perf(win): use ThreadPool for video rendering & EGL/ANGLE usage synchronization

alexmercerind:
~~Now Flutter UI feels much snappier than before & no longer feels "slowed down" when one or more videos are visible.~~

~~Technically, the correct place to do all the video rendering (& thus resize) etc. is on Flutter's render thread itself (exposed as callback in flutter::GpuSurfaceTexture & flutter::PixelBufferTexture). However, this slows down the UI too much. So, I decided to create our own ThreadPool & use it for all the video rendering purposes through one single thread.~~

All the operations involving ANGLE or EGL or libmpv must be performed on same single thread to prevent any race conditions or invalid ANGLE usage. Not doing so results in access violations & crashes (#17).

The following operations are performed through the ThreadPool:
* Rendering of video frame i.e. mpv_render_context_render is called through the ThreadPool after being notified by mpv_render_context_set_update_callback.
* Creation / Disposal of new VideoOutput i.e. VideoOutputManager::Create & VideoOutputManager::Dispose are called through the ThreadPool. Thus, mpv_render_context_create is called & instantiation of a new ANGLESurfaceManager is done through ThreadPool (in VideoOutput constructor).
* Resizing of ANGLESurfaceManager & creation of newly sized Flutter textures (flutter::GpuSurfaceTexture & flutter::PixelBufferTexture).

All of these involve ANGLE or OpenGL context etc. etc.

Creating a ThreadPool with maximum number of worker threads as 1, ensures that all the ANGLE, EGL & libmpv operations are performed on a single thread **orderly**. This also makes usage of any std::mutex unnecessary (for the good).

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,13 +423,11 @@ classDiagram
   }
 
   class VideoOutputManager {
-    +Create(handle: int, width: optional<int>, height: optional<int>)
+    +Create(handle: int, width: optional<int>, height: optional<int>, texture_update_callback: std::function)
     +Dispose(handle: int)
 
-    -std::mutex mutex_
     -std::mutex render_mutex_
     -flutter::PluginRegistrarWindows registrar_
-    -std::unique_ptr<MethodChannel> channel_
     -std::unordered_map<int64_t, std::unique_ptr<VideoOutput>> video_outputs_
   }
 
--- a/README.md
+++ b/README.md
@@ -245,7 +245,6 @@ classDiagram
   }
 
   class PlayerConfiguration {
-    + bool texture
     + bool osc
     + String vo
     + String title
@@ -413,7 +412,9 @@ classDiagram
 
   MediaKitVideoPlugin "1" *-- "1" VideoOutputManager: Create VideoOutput(s) with VideoOutputManager for handle passed through platform channel
   VideoOutputManager "1" *-- "*" VideoOutput
-  VideoOutput "1" *-- "1" ANGLESurfaceManager: Only for H/W accelerated rendering.
+  VideoOutputManager "1" *-- "1" ThreadPool: 
+  VideoOutput "1" o-- "1" ThreadPool: Post creation, resize & render etc. tasks involving EGL to ensure synchronous EGL/ANGLE usage across multiple VideoOutput(s)
+  VideoOutput "1" *-- "1" ANGLESurfaceManager: Only for H/W accelerated rendering
 
   class MediaKitVideoPlugin {
     -flutter::PluginRegistrarWindows registrar_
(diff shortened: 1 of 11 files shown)


alexmercerind:
Theory aside...

<hr>

From my personal experience, earlier I could notice the ripple effect in various buttons etc. kind of "lag". Now that no longer happens. Flutter UI in general feels much smoother when video(s) are being rendered.

Any flickers (#10) or crashes (#17) after this PR?

Download Build: https://github.com/alexmercerind/media_kit/suites/10446789924/artifacts/516952921

Quackdoc:
Flashing is really bad on polaris rx580, on both the artifact and a clean build

alexmercerind:
> Flashing is really bad on polaris rx580, on both the artifact and a clean build

I'm curious, how does the last "Stress Test" screen behave?

<hr>

Both, artifact & local builds should be identical. Feel free to just test the artifacts.

alexmercerind:
You mean local files & other files are flickering, but this video in specific is flickering _relatively more_?

<hr>

This is flickering for me too!!! Finally I can test out with something.

Quackdoc:
correct, it is quite possibly because MPV is tonemapping it too, however I couldn't say if it was just luck

alexmercerind:
It shouldn't flicker anymore. Primarily because the mechanism is quite same as before.

Waiting for your approval.

Download Artifact: https://github.com/alexmercerind/media_kit/suites/10456379301/artifacts/517683147

<hr>

**What's the issue you tried to fix?**

In current setup (even before this pull request), video frames rendering is "sandwich-ed" into Flutter's own rendering loop. This means a video frame is rendered in-between Flutter's own rendering. If multiple videos are being displayed, each of those render one-by-one between Flutter's own frames.

In those "sandwich-ed" operations, few other operations besides rendering are also inserted (based on requirements) which include creation of new video surface/texture, resize of video surface/texture. This is because all of these must be called one-by-one (_see last sentence of this section_).

This syncs everything correctly & eliminates any flickers. However, performance is kind-of not that good (from one point of view) because Flutter's own UI gets "slowed" due to latency caused by videos' rendering.

AFAIK, OpenGL is meant to be used on single thread as-well. When it's not, it crashes (#17). So maybe there's no way to achieve what I'm trying to do. Maybe rendering one-by-one is the necessary evil.

<hr>

**Why do I still wanna merge this?**

Because this PR cleans up that synchronization process regardless. Now all ANGLE/EGL operations are "piped" through a `ThreadPool` (single thread). When a `Render` thread is piped however, we wait for it (so that it doesn't flicker).

Quackdoc:
I cant review the code or comments due to being extremely busy with family matters, however I have been able to abuse the program quite badly, and it preformed phenomenally

I tested 8k HDR video playback, while running a MSI kombustor stress test in the foreground, not a single flicker, so this is even improved upon in that regard. performance is also markedly better even in debug mode, resizing the window is smooth too.

LGTM

PR #29, opened 2023-01-23
perf(win): improve rendering synchronization & support stable channel

alexmercerind:
## Introduction

Now two instances of `ID3D11Texture2D` are created inside `ANGLESurfaceManager`. First one for drawing onto using ANGLE/EGL (libmpv) & second one which gets copied from the first one upon every request. The `ID3D11Texture2D` requested by the public API & gets copied to right before every render request (ANGLESurfaceManager::Read). It's shared HANDLE is exposed in the `ANGLESurfaceManager::handle` getter. 

`ANGLESurfaceManager::Draw` & `ANGLESurfaceManager::Read` abstract the process of EGL context acquiring & rendering. A single mutex is also involved.

This removes any possible races & synchronization issues between libmpv & Flutter _i.e._ flickers (#10) without needing any redundant `glFinish` calls (#11).

- `ANGLESurfaceManager::Draw`:
  - Call ANGLE/EGL/libmpv APIs in it's callback & perform rendering onto internal `ID3D11Texture2D`.
- `ANGLESurfaceManager::Read`:
  - Access the rendered content [after copying the internal texture to a separate public texture](https://learn.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-copyresource) (& then reading it).

<hr>

## Drawbacks

- Two copies of `ID3D11Texture2D`s are created for each video output.
- A copy (though, GPU to GPU) is involved. Technically there should be a slight affect in performance.

Personally, I can't notice any difference performance-wise even on my slow machine, so I believe it is negligible. Since, we were using a long trail of `glFinish` already, I think this is already better & more stable solution for the future of the project.

## Advantages

- Better synchronization between libmpv & Flutter rendering. 
- No multiple `glFinish` calls between each frame.

As last pull request (#27) introduced a new synchronization mechanism i.e. `ThreadPool`, a large number of issues were solved & major part of implementation was decoupled from Flutter's own rendering cycle. So, I could make [package:media_kit](https://github.com/alexmercerind/media_kit) work with stable channel of Flutter 3.3.x (stable).

- Support for Flutter 3.3.x (stable).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
           submodules: "true"
       - uses: subosito/flutter-action@v1
         with:
-          channel: "master"
+          channel: "stable"
       - run: flutter pub get
       - run: flutter build windows --verbose
       - uses: actions/upload-artifact@v1
--- a/README.md
+++ b/README.md
@@ -412,8 +412,8 @@ classDiagram
 
   MediaKitVideoPlugin "1" *-- "1" VideoOutputManager: Create VideoOutput(s) with VideoOutputManager for handle passed through platform channel
   VideoOutputManager "1" *-- "*" VideoOutput
-  VideoOutputManager "1" *-- "1" ThreadPool:
-  VideoOutput "1" o-- "1" ThreadPool: Post creation, resize & render etc. tasks involving EGL to ensure synchronous EGL/ANGLE usage across multiple VideoOutput(s)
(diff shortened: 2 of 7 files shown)


alexmercerind:
Any flickers?

https://github.com/alexmercerind/media_kit/suites/10527445482/artifacts/523078821

Quackdoc:
Passes the 8k test with flying colours

PR #34, opened 2023-02-02
feat: linux video embedding

alexmercerind:
All the cases are passing.

### Overview

As far as I can see, resource usage is quite minimal. I can play 4K videos easily.

All this rendering is being done using https://github.com/flutter/engine/pull/24916.

### Comparison with Windows

One good thing that I've done better on Windows is that [it renders video on another thread & only copies already rendered texture to another texture on render thread (for access by Flutter)](https://github.com/alexmercerind/media_kit/blob/487ce8c690c0abc48937e70e79fadfc623f9dff8/media_kit_video/windows/angle_surface_manager.cc#L221-L226), so Flutter UI doesn't freeze or slow down because we are performing expensive `mpv_render_context_render` on another thread. On Linux however, all the rendering is being done on render thread (theoretically correct) & I can't find any other way to call `mpv_render_context_render` & GL friends on another thread (I don't think it's possible). In general, UI is feeling smoother on Windows when one or more videos are playing.

In theory, this is more efficient than Windows (_still very-very good & as I said above more "smooth"_). AFAIK, render API from libmpv copies frames to CPU internally on Windows (I can't tell how & to what extent, mpv developers know better). Intermediate ANGLE / EGL involvement is the main cause, which causes libmpv to switch to `*-copy` `hwdec` on Windows. A direct D3D / swapchain API (https://github.com/mpv-player/mpv/issues/5979) will be **even better** & 100 % GPU based. In my personal tests however, `*-copy` `hwdec`s actually make mpv perform better than non-`*-copy` `hwdec`s, so maybe we don't need that.

<hr>

Screenshot from 2023-02-04 01 12 26
Screenshot from 2023-02-04 01 12 56
Screenshot from 2023-02-04 01 13 31
Screenshot from 2023-02-04 01 14 16


alexmercerind:
The API & scaffolding is done. Only implementation is left.

The code is very standard & uses GObject & Glib equivalents. I'm very satisfied. Even though both Windows & Linux implementations are in C++, Windows implementation uses standard C++ (`std::`) & Win32, while this is GObject based C/C++. This includes both types & memory management e.g. `GHashTable`/`std::unordered_map`, `g_autoptr`/`std::unique_ptr` etc.

alexmercerind:
Something finally worked!



### Thoughts

As far as I can see, resource usage is quite minimal. Above is a 4K 30 FPS video. Rest, it also depends upon the file codec/format which is being decoded by libmpv. In some cases, a 4K 60 FPS is taking less CPU than 4K 30 FPS.

Flutter itself suffers from ton of issues at the moment which are taking a toll on CPU. A major one: https://github.com/flutter/flutter/issues/88996.
I think this is only X11 specific. I'm also on X11. Wayland will perform better. Still, this is looking good to me in either case.

All this rendering is being done using https://github.com/flutter/engine/pull/24916.

<hr>

### Comparison with Windows

In theory, this is more efficient than Windows (_still very-very good_). AFAIK, render API from libmpv copies frames to CPU internally on Windows (I can't tell how & to what extent, mpv developers know better). Intermediate ANGLE / EGL involvement is the main cause, which causes libmpv to switch to `*-copy` `hwdec` on Windows. A direct D3D / swapchain API (https://github.com/mpv-player/mpv/issues/5979) will be **even better** & 100 % GPU based. In my personal tests however, `*-copy` `hwdec`s actually make mpv perform better than non-`*-copy` `hwdec`s, so maybe we don't need that.

alexmercerind:
Everything is working correctly!!!

I've lost my mind. Something is wrong within Flutter or Skia or the very specific scenario that's taking place.

But the default back button animation in `Navigator` i.e. `ZoomPageTransitionsBuilder` throws the UI into deadlock.
I changed it to `OpenUpwardsPageTransitionsBuilder` as follows & nothing freezes!

```dart
    return MaterialApp(
      theme: ThemeData(
        primarySwatch: Colors.blue,
        fontFamily: Platform.isLinux ? 'Inter' : null,
        pageTransitionsTheme: const PageTransitionsTheme(
          builders: {
            TargetPlatform.android: OpenUpwardsPageTransitionsBuilder(),
            TargetPlatform.iOS: OpenUpwardsPageTransitionsBuilder(),
            TargetPlatform.linux: OpenUpwardsPageTransitionsBuilder(),
            TargetPlatform.macOS: OpenUpwardsPageTransitionsBuilder(),
            TargetPlatform.windows: OpenUpwardsPageTransitionsBuilder(),
          },
        ),
      ),
      home: const HomeScreen(),
    );
```

alexmercerind:
FYI, you can clone & compile this branch & it will work now.

alexmercerind:
Added fallback pixel buffer based software rendering, and it's feeling "smoother" than hardware rendering.

It because I'm rendering into pixel buffer on another thread. It's a pity that I can't use OpenGL on other threads (for hardware rendering) like I could do with ANGLE on Windows. Or is there a way?

alexmercerind:
> Added fallback pixel buffer based software rendering, and it's feeling "smoother" than hardware rendering.
> 
> It because I'm rendering into pixel buffer on another thread. It's a pity that I can't use OpenGL on other threads (for hardware rendering) like I could do with ANGLE on Windows. Or is there a way?

Okay! I could make rendering work from another thread. Now I can make some more performance improvements & mark this pull-request ready for merge.

alexmercerind:
MERGE!

<hr>

BTW, this PR also adds a new API to switch between turn on/off hardware acceleration (on by default). 
```dart
final controller = await VideoController.create(
  player.handle,
  enableHardwareAcceleration: false,      // default: true
);
```

I'll make this argument work on Windows with a follow-up PR. You guys _really_ should not change it though. There isn't any real benefit in turning it off. Use hardware acceleration, reduce CPU usage, save battery life & improve performance.

Feichtmeier:
Great job!

srix55:
Awesome... you are. Thanks. Will test it soon.

alexmercerind:
> Awesome... you are. Thanks. Will test it soon.

Thanks for appreciation!

There are pre-built binaries of the test application as well (scroll down):

https://github.com/alexmercerind/media_kit/actions/runs/4092963785

Just make sure you have `libmpv2` package, if you're on Ubuntu or Debian.

srix55:
Looks good. I am on Fedora 36. I had to install `mpv-libs` and `mpv-libs-devel` for the linux test app [binary](https://github.com/alexmercerind/media_kit/suites/10781076409/artifacts/541799578) to work. Here are the results.

[Screencast from 07-02-23 03:54:43 PM IST.webm](https://user-images.githubusercontent.com/6329365/217219131-44c45102-c23d-46f0-8315-6e1e8f60b51a.webm)

A few stutters, especially in tabbed simultaneous videos. Mine is an intel i5, 6th gen processor, with an nvidia 2gb gt 1030 card. Will do more test runs.

alexmercerind:
> A few stutters, especially in tabbed simultaneous videos.
> Mine is an intel i5, 6th gen processor, with an nvidia 2gb gt 1030 card. Will do more test runs.

I have discussed the issue in the PR comment & few individual comments above.

GTK allows usage of OpenGL only on render thread. Since Flutter is also rendered using GL, it gets slowed down because mpv is also rendering on the same thread. AFAIK there's no way around it. Still, I'm not much knowledgeable about GTK & using OpenGL on it.

You will notice substantial less "jitter" when forcing enableHardwareAcceleration: false, because it renders on separate (non-UI) threads (_because sharing CPU RAM i.e. pixel buffer can be done across threads without platform / GTK / GL limitations_). This comes at the cost of higher CPU usage & lesser battery life.

More details are already mentioned above. I'm welcomed to any improvements, this is my best GTK knowledge.

alexmercerind:
Was switching to GTK really a good decision? Sometimes I wonder GLFW was better than this.

srix55:
It's impressive that it can handle 4 players side by side on the UI thread. Will run some cases with hardware acceleration turned off.

talksik:
Hey @alexmercerind, can you please help me out? I am trying to do something similar with my own custom flutter video capture + render plugin.

1. I kinda understand the memory inefficiency with using `flutter::PixelBufferTexture`. However, I don't understand the hardware acceleration part. On Linux desktop, skia will use openGL to render. If my openGL is using intel graphics driver, then it should be using hardware acceleration. Thus, even though using pixel buffer is inefficient with the memory roundtrips, it is still hardware accelerated rendering, correct?
2. I am super noobie with linux graphics, can you please explain how this pull request achieves the goal? Does [flutter/engine#24916](https://github.com/flutter/engine/pull/24916) enable external textures, and so you were able to find a way for `MPV` to output an openGL texture and give that to flutter? Is that the overall gist? 

Would really appreciate it as I have been trying to dig around to understand these concepts but just stuck. Thank you!!!

alexmercerind:
@talksik 

1. 

flutter:PixelBufferTexture is available on Microsoft Windows, FlPixelBufferTexture is GNU/Linux equivalent.

There are two parts to video playback, decoding & rendering. Although it depends upon the underlying library that you're using (libmpv here), you can assume that decoding will be hardware accelerated even if you use pixel-buffer. The most efficient situation is when both decoding & rendering is hardware-accelerated/GPU-based (without any RAM copy).

You can refer to media_kit's architecture & implementation details in the README, where everything was documented during early stage development. Alternatively, code can itself answer your questions.

> It is still hardware accelerated rendering, correct? 

Only decoding. You'll still see noticeable CPU usage.

media_kit offers both types of video rendering (with automatic selection or user selection): GL based (H/W) & pixel-buffer (S/W) based. Different methods are used on all platforms to translate/host GL. In future, a direct D3D rendering will be added on Windows, ref: #296.

2. 

libmpv provides both GL-based & pixel-buffer-based rendering APIs.

talksik:
Thank you so much @alexmercerind !!! This is very helpful.

I appreciate you breaking down the difference between decoding and rendering. Since I am doing stuff without mpv, I have a gstreamer pipeline that uses hardware accelerated decoders, but need to figure out how to do hardware accelerated rendering. I am going to ask the folks [here](https://github.com/sony/flutter-elinux-plugins/pull/67) here if that is using hardware acceleration, but need to dig in further how to use gstreamer optimally with flutter `TextureRegistry` for someone who doesn't know openGL that well.

Issue #40, opened 2023-02-05
[Bug] 8 or more Player(s) freeze the application


alexmercerind:
Filed https://github.com/dart-lang/sdk/issues/51254 for now.

Let's see what should be done in regard to this. It is quite important to fix this. Many users of [package:media_kit](https://github.com/alexmercerind/media_kit) are definitely playing more than 8 videos at once. Reaching that level of stability/performance is certainly a goal for project.

8 is definitely not a less number. Most applications are never going to play this number of videos simultaneously.

<hr>

You can +1 this issue comment. Working on this will require substantial time.

alexmercerind:
I will work on this as part of #41.

Issue #20, opened 2023-01-14
[Enhancement] custom libmpv shared libraries


alexmercerind:
We need own libmpv builds for:
- Allowing usage of `package:media_kit` in proprietary/commercial applications (by using LGPL compliant libmpv).
- Reducing the final bundle size of the resulting application using `package:media_kit` as it's dependency.
- Keeping specialized libmpv binaries available for download (e.g. GitHub releases) to to build projects having `package:media_kit`.

With the highest priority being first:

### Windows [Done]

[media_kit_core_audio](https://github.com/alexmercerind/media_kit/tree/main/media_kit_core_audio) bundles audio-only `libmpv-2.dll` on Windows.
[media_kit_core_video](https://github.com/alexmercerind/media_kit/tree/main/media_kit_core_video) bundles minimal video+audio `libmpv-2.dll` on Windows.

Both are LGPL compliant & allow commercial usage.

__References:__

- https://github.com/harmonoid/harmonoid-custom-codec was used to build the binary. no longer works, needs updates.
- Forking [this (internally clones the second)](https://github.com/zhongfly/mpv-winbuild) or [this](https://github.com/shinchiro/mpv-winbuild-cmake), disabling GPL modules & compiling LGPL FFmpeg should work.
- Further information is already present at https://github.com/harmonoid/harmonoid-custom-codec.
- One [might have to enable logging with `continue-on-error: true` to deal with linking errors etc. etc.](https://github.com/harmonoid/harmonoid-custom-codec/blob/11f6ccb87f7abd4442698cc8f59ef0c898bd118f/.github/workflows/ci.yml#L55-L63).

### Linux [Done]

`package:media_kit` doesn't bundle anything related to libmpv / FFmpeg etc. etc. **It is not feasible either**.

- On Linux, distribution specific packages are the "standard".
- Distribution specific packages are expected to be installed by the users themselves before installing application package.
- Since, any commercial / proprietary application won't distribute these, it's not important for these to be LGPL etc.
- In case of an alternative distribution methodology e.g. Snapcraft, Flathub etc., it is responsibility of the `package:media_kit` dependents to handle it themselves. Few references in regards to this may be located [here](https://github.com/ozmartian/vidcutter/blob/master/_packaging/flatpak/com.ozmartians.VidCutter.json), where libmpv, FFmpeg etc. are "bundled" into Flatpak package.
  
 ### macOS [WIP]
 
[...]

PR #44, opened 2023-02-19
add macos audio & video support

### LINKED ISSUE #43: [Enhancement] macOS support
I can't find a suitable library on macOS.

How long can I expect to support macOS?

birros:
:tada: Fix #43

> :warning: If you experience lag when playing heavy videos (4K, 8K…), consider reducing the rendering resolution by adding `width` and `height` with the controller.

## What works in this pull request

- [X] Software rendering, directly in a pixel buffer (RAM)
- [X] "Partial" hardware rendering, using OpenGL but with two drawbacks:
    1. We are using the Apple version of OpenGL, which is not future-proof since Apple has deprecated its support
    2. We are using a pixel buffer to communicate with Flutter, which has a theoretical impact on performance and power consumption, although in practice this is reasonable for moderate use

## What remains to be done in this pull request

- [x] Fix dylibs's install
- [x] Set up a CI build of libmpv for macOS, in another repo, to not depend on **IINA** anymore
- [x] Improve and check the code for potential memory leaks, and improve the "quality" of the code
- [x] Add "some" documentation to make it a bit more digestible for potential contributors

## For a future pull request

- "Full" hardware rendering, i.e. no pixel buffer and deprecated OpenGL, which means rendering in a Metal texture, directly used by Flutter. The choice of **ANGLE** is relevant, as it allows to convert **OpenGL ES** calls to **Metal**, but it has a significant drawback. When mpv uses **OpenGL ES**, it is forced to use a buffer (RAM) when decoding the video (i.e. the `hwdec` chosen is of type `*-copy`). Both "partial" and "full" methods have their advantages and disadvantages, neither will be perfect, although the latter is much more future proof. Another possibility is Vulkan, the SDK seems to render directly to Metal using MoltenVK


alexmercerind:
First of all, this is awesome. 👏👏👏

I serve own libmpv builds on Windows for both media_kit_libs_windows_video & media_kit_libs_windows_audio:

- https://github.com/alexmercerind/mpv-win32-build/releases
- https://github.com/harmonoid/mpv-win32-build/releases

The later has lesser number of decoders & strips video related libraries, thus bundle size is less. I think we might need same for macOS. This will also help us allow the commercial usage of the package. See #20.

<hr>

Yeah! Only Windows version is multi-threaded currently & makes everything appear way more smoother because mpv & Flutter rendering is separated. I wonder if same can be done here or on GNU/Linux.

I can VNC into a macOS machine for testing & improving things, let's see how good that goes.

damywise:
Awesome!
I just tested this on my M1 Macbook Pro
Some reviews:
1. `Running pod install...` is taking too much time. Like ~5-7 minutes.
2.  This error when first ever `flutter run -d macos`.
```sh
#import "Headers/media_kit_video-Swift.h"
        ^
/Users/XXXXXXXXXXXXXXXXXXXXXX/Documents/GitHub/media_kit/media_kit_test/build/macos/Build/Products/Debug/media_kit_video/media_kit_video.framework/Headers/media_kit_video-Swift.h:270:31: warning: 'objc_ownership' only applies to Objective-C object or block pointer types; type here is 'CVPixelBufferRef' (aka 'struct __CVBuffer *')
- (CVPixelBufferRef _Nullable __unsafe_unretained)copyPixelBuffer SWIFT_WARN_UNUSED_RESULT;
                              ^
```
3. On 1 player 1 video, GPU is maxed out on ~8% while CPU maxed out at ~35%
4. On 1 players 4 videos, GPU is maxed out on ~16% while CPU maxed out at ~37%
5. On 2 players 2 videos,, GPU is maxed out on ~22% while CPU maxed out at ~50%
6. On tabs, it just crashes when I changed tab.
7. On stress test, GPU is maxed out on ~26%, CPU maxed out at 50%, Is 
8. On all situations, video always flickers heavily. Audio is pretty much fine
9. When video is playing, UI is incredibly laggy.
10. Lots of the time, GPU is not used at all.

birros:
@damywise: Thx you for your feedback 👍 

> 1. On first run, `Running pod install...` is taking an incredibly long time. Like ~5-7 minutes.

This seems you are using **rosetta**, it should not exceed 2 minutes without rosetta (like on my old x86 machine).

> 2. This error when first ever `flutter run -d macos`.
> 
> ```shell
> #import "Headers/media_kit_video-Swift.h"
>         ^
> /Users/XXXXXXXXXXXXXXXXXXXXXX/Documents/GitHub/media_kit/media_kit_test/build/macos/Build/Products/Debug/media_kit_video/media_kit_video.framework/Headers/media_kit_video-Swift.h:270:31: warning: 'objc_ownership' only applies to Objective-C object or block pointer types; type here is 'CVPixelBufferRef' (aka 'struct __CVBuffer *')
> - (CVPixelBufferRef _Nullable __unsafe_unretained)copyPixelBuffer SWIFT_WARN_UNUSED_RESULT;
>                               ^
> ```

This is a warning, not an error. I also have it, I followed the official recommendation to remove this warning by adding the `@discardableResult` attribute to the `copyPixelBuffer` method, without success. So I removed this attribute, as it had no effect. I doubt that this has any impact on the stability or performance of the created app, so it should be ignored.

> 3. On 1 player 1 video, GPU is maxed out on ~8% while CPU maxed out at ~35%
> 4. On 1 player 4 videos, GPU is maxed out on ~16% while CPU maxed out at ~37%
> 5. On 2 players 2 videos,, GPU is maxed out on ~22% while CPU maxed out at ~50%
> 6. On tabs, it just crashes when I changed tab.
> 7. On stress test, GPU is maxed out on ~26%, CPU maxed out at 50%, Is
> 8. On all situations, video always flickers heavily. Audio is pretty much fine
> 9. When video is playing, UI is incredibly laggy.

rosetta again, IINA dylibs are universal, which means they include arm64 support, so this comes from using the rosetta version of Flutter. Try upgrading to an M1 version of Flutter, perhaps on the beta or master channel.

> 10. Lots of the time, GPU is not used at all.

Do you mean that the GPU is not used when idle? Which is normal.

damywise:
> Do you mean that the GPU is not used when idle? Which is normal.
I think I meant that when the video is playing, the GPU usage jumps between 0 and 9 on Activity Monitor. But that's probably a mistake on my side.

I tried upgrading to master. It's so much worse now lol.
Previously it was building for arm64, now it's building for x86_64.

CPU jumps to 71%. GPU is still ~8% or 0%.
It feels like GPU isn't used by mpv at all, only by Flutter.

birros:
> > Do you mean that the GPU is not used when idle? Which is normal.
> 
> I think I meant that when the video is playing, the GPU usage jumps between 0 and 9 on Activity Monitor. But that's probably a mistake on my side.
> 
> I tried upgrading to master. It's so much worse now lol. Previously it was building for arm64, now it's building for x86_64.
> 
> CPU jumps to 71%. GPU is still ~8% or 0%. It feels like GPU isn't used by mpv at all, only by Flutter.

Sorry for making you switch to master. If the build is arm64 and the libs are universal, this is caused by something else...

I think you are right, the GPU seems not to be used by mpv on your M1, which is not the case on x86.

Does IINA work well on an M1?

damywise:
> Does IINA work well on an M1?

Yep. I've been using IINA for quite some time.

birros:
@damywise I may have a clue, it may not be the OpenGL rendering, but rather the video decoding.

> 8. On all situations, video always flickers heavily. Audio is pretty much fine

I managed to reproduce this behaviour by setting `hwdec` with `videotoolbox-copy`, setting the render size to 1440x900, and playing a 4K video from a URI.

When I use `videotoolbox`, the rendering goes back to normal, probably the `auto` mode in your case does not select the good decoding mode.

If you have time, I would be grateful if you could test which value for `hwdec` is most relevant in your case, if it makes a difference. I don't know how to get which "auto" mode is selected by mpv.

- You can find the possible values here: [mpv.io/manual/master/#options-hwdec](https://mpv.io/manual/master/#options-hwdec)
- The corresponding line to change: [media_kit_video/macos/Classes/VideoOutput.swift#L84](https://github.com/birros/media_kit/blob/7311039/media_kit_video/macos/Classes/VideoOutput.swift#L84)

I think you should test `videotoolbox` first, then, if that doesn't work, try the `auto(-*)` values. Some of them seem to be useful only on other platforms.

damywise:
I tried changing it to `videotoolbox`, `videotoolbox-copy`, `auto-safe`, `auto-copy`. I don't think there's significant changes, if any.	

> If you don't have time, or it doesn't work, can you at least give us the logs of this command:
>
>```sh
>$ mpv -v ~/Downloads/BigBuckBunny.mp4     
>```

Here you go:
```sh

[cplayer] Command line options: '-v' './video_1.mp4'
[cplayer] mpv 0.35.0 Copyright © 2000-2022 mpv/MPlayer/mplayer2 projects
[cplayer]  built on Sat Nov 12 14:47:29 UTC 2022
[cplayer] FFmpeg library versions:
[cplayer]    libavutil       57.28.100
[cplayer]    libavcodec      59.37.100
[cplayer]    libavformat     59.27.100
[cplayer]    libswscale      6.7.100
[cplayer]    libavfilter     8.44.100
[cplayer]    libswresample   4.7.100
[cplayer] FFmpeg version: 5.1.2
[cplayer] 
[cplayer] Configuration: waf configure --prefix=/opt/homebrew/Cellar/mpv/0.35.0 --enable-html-build --enable-javascript --enable-libmpv-shared --enable-lua --enable-libarchive --enable-uchardet --confdir=/opt/homebrew/etc/mpv --datadir=/opt/homebrew/Cellar/mpv/0.35.0/share/mpv --mandir=/opt/homebrew/Cellar/mpv/0.35.0/share/man --docdir=/opt/homebrew/Cellar/mpv/0.35.0/share/doc/mpv --zshdir=/opt/homebrew/Cellar/mpv/0.35.0/share/zsh/site-functions --lua=luajit
[cplayer] List of enabled features: asm av-channel-layout bsd-fstatfs build-date cocoa coreaudio cplayer cplugins debug-build ffmpeg gl gl-cocoa glob glob-posix gpl iconv javascript jpeg jpegxl lcms2 libarchive libass libavdevice libbluray libdl libm libmpv-shared lua luajit macos-10-11-features macos-10-12-2-features macos-10-14-features macos-cocoa-cb macos-media-player macos-touchbar optimize osx-thread-name plain-gl posix posix-or-mingw pthreads rubberband rubberband-3 stdatomic swift uchardet vapoursynth vector videotoolbox-gl videotoolbox-hwaccel zimg zlib
[cplayer] Reading config file /opt/homebrew/etc/mpv/encoding-profiles.conf
[cplayer] Applying profile 'default'...
[cplayer] Setting option 'v' = '' (flags = 8)
[osx/cocoacb] Created CGL pixel format with attributes: kCGLPFAOpenGLProfile, kCGLOGLPVersion_3_2_Core, kCGLPFAAccelerated, kCGLPFADoubleBuffer, kCGLPFAColorSize, 64, kCGLPFAColorFloat, kCGLPFABackingStore, kCGLPFAAllowOfflineRenderers, kCGLPFASupportsAutomaticGraphicsSwitching, 0
[libmpv_render] GL_VERSION='4.1 Metal - 76.3'
[libmpv_render] Detected desktop OpenGL 4.1.
[libmpv_render] GL_VENDOR='Apple'
[libmpv_render] GL_RENDERER='Apple M1'
[libmpv_render] GL_SHADING_LANGUAGE_VERSION='4.10'
[libmpv_render] Loaded extension GL_APPLE_rgb_422.
[libmpv_render] GL_*_swap_control extension missing.
[libmpv_render] Testing FBO format rgba16f
[libmpv_render] Using FBO format rgba16f.
[libmpv_render] Disabling HDR peak computation (one or more of the following is not supported: compute shaders=0, SSBO=0).
[libmpv_render] No advanced processing required. Enabling dumb mode.
[libmpv_render] Loading hwdec driver 'videotoolbox'
[cplayer] Waiting for scripts...
[osd/libass] libass API version: 0x1700000
[osd/libass] libass source: tarball: 0.17.0
[osd/libass] Shaper: FriBidi 1.0.12 (SIMPLE) HarfBuzz-ng 7.0.0 (COMPLEX)
[osd/libass] Setting up fonts...
[osd/libass] Using font provider coretext
[osd/libass] Done.
[cplayer] Set property: shared-script-properties -> 1
[cplayer] Set property: shared-script-properties -> 1
[cplayer] Done loading scripts.
[cplayer] Running hook: ytdl_hook/on_load
[ytdl_hook] ytdl:// hook 
[ytdl_hook] not a ytdl:// url 
[bdmv/bluray] Opening ./video_1.mp4
[file] Opening ./video_1.mp4
[demux] Trying demuxers for level=normal.
[cplayer] Set property: shared-script-properties -> 1
[osd/libass] libass API version: 0x1700000
[osd/libass] libass source: tarball: 0.17.0
[osd/libass] Shaper: FriBidi 1.0.12 (SIMPLE) HarfBuzz-ng 7.0.0 (COMPLEX)
[osd/libass] Setting up fonts...
[osd/libass] Using font provider coretext
[osd/libass] Done.
[cplayer] Set property: shared-script-properties -> 1
[cplayer] Set property: shared-script-properties -> 1
[lavf] Found 'mov,mp4,m4a,3gp,3g2,mj2' at score=100 size=2048.
[demux] Detected file format: mov,mp4,m4a,3gp,3g2,mj2 (libavformat)
[cplayer] Opening done: ./video_1.mp4
[find_files] Loading external files in ./
[cplayer] Running hook: ytdl_hook/on_preloaded
[lavf] select track 0
[lavf] select track 1
[cplayer]  (+) Video --vid=1 (*) (h264 1280x720 24.000fps)
[cplayer]  (+) Audio --aid=1 (*) (aac 2ch 44100Hz)
[osx/cocoacb] Can't find an ambient light sensor
[vo/libmpv] Assuming 60.002400 FPS for display sync.
[vd] Container reported FPS: 24.000000
[vd] Codec list:
[vd]     h264 - H.264 / AVC / MPEG-4 AVC / MPEG-4 part 10
[vd] Opening decoder h264
[vd] No hardware decoding requested.
[vd] Using software decoding.
[vd] Detected 8 logical cores.
[vd] Requesting 9 threads for decoding.
[vd] Selected codec: h264 (H.264 / AVC / MPEG-4 AVC / MPEG-4 part 10)
[vf] User filter list:
[vf]   (empty)
[ad] Codec list:
[ad]     aac - AAC (Advanced Audio Coding)
[ad]     aac_fixed (aac) - AAC (Advanced Audio Coding)
[ad]     aac_at (aac) - aac (AudioToolbox)
[ad] Opening decoder aac
[ad] Requesting 1 threads for decoding.
[ad] Selected codec: aac (AAC (Advanced Audio Coding))
[af] User filter list:
[af]   (empty)
[cplayer] Starting playback...
[vd] DR failed - disabling.
[vd] Using software decoding.
[vd] Decoder format: 1280x720 [0:1] yuv420p auto/auto/auto/auto/auto CL=mpeg2/4/h264
[vf] [in] 1280x720 yuv420p bt.709/bt.709/bt.1886/limited/display SP=1.000000 CL=mpeg2/4/h264
[vf] [userdeint] 1280x720 yuv420p bt.709/bt.709/bt.1886/limited/display SP=1.000000 CL=mpeg2/4/h264
[vf] [userdeint] (disabled)
[vf] [autorotate] 1280x720 yuv420p bt.709/bt.709/bt.1886/limited/display SP=1.000000 CL=mpeg2/4/h264
[vf] [autorotate] (disabled)
[vf] [convert] 1280x720 yuv420p bt.709/bt.709/bt.1886/limited/display SP=1.000000 CL=mpeg2/4/h264
[vf] [convert] (disabled)
[vf] [out] 1280x720 yuv420p bt.709/bt.709/bt.1886/limited/display SP=1.000000 CL=mpeg2/4/h264
[af] [in] 44100Hz stereo 2ch floatp
[af] [userspeed] 44100Hz stereo 2ch floatp
[af] [userspeed] (disabled)
[af] [convert] 44100Hz stereo 2ch floatp
[ao] Trying audio driver 'coreaudio'
[ao/coreaudio] requested format: 44100 Hz, stereo channels, floatp
[ao/coreaudio] selected audio output device: MacBook Pro Speakers (50)
[ao/coreaudio] input channel layout:
[ao/coreaudio] layout: tag: <0>, bitmap: <0>, descriptions <2>
[ao/coreaudio]  - description 0: label <4294967295, 64>,  flags: <0>, coords: <0.000000, 0.000000, 0.000000>
[ao/coreaudio]  - description 1: label <4294967295, 64>,  flags: <0>, coords: <0.000000, 0.000000, 0.000000>
[ao/coreaudio] mp chmap: unknown2
[ao/coreaudio] input channel layout:
[ao/coreaudio] layout: tag: <0>, bitmap: <0>, descriptions <2>
[ao/coreaudio]  - description 0: label <1, 0>,  flags: <0>, coords: <0.000000, 0.000000, 0.000000>
[ao/coreaudio]  - description 1: label <2, 1>,  flags: <0>, coords: <0.000000, 0.000000, 0.000000>
[ao/coreaudio] mp chmap: stereo
[ao/coreaudio] Channel layouts:
[ao/coreaudio]  - stereo
[ao/coreaudio]  - mono
[ao/coreaudio] result: stereo
[ao/coreaudio] Latency property cntl: 66 frames
[ao/coreaudio] Latency property zisf: 512 frames
[ao/coreaudio] Latency property tfas: 113 frames
[ao/coreaudio] audiounit latency [us]: 362
[ao/coreaudio] device latency [us]: 15668
[ao/coreaudio] using soft-buffer of 8820 samples.
[cplayer] AO: [coreaudio] 44100Hz stereo 2ch floatp
[cplayer] AO: Description: CoreAudio AudioUnit
[af] [convert] (disabled)
[af] [out] 44100Hz stereo 2ch floatp
[cplayer] VO: [libmpv] 1280x720 yuv420p
[cplayer] VO: Description: render API for libmpv
[vo/libmpv] reconfig to 1280x720 yuv420p bt.709/bt.709/bt.1886/limited/display SP=1.000000 CL=mpeg2/4/h264
[libmpv_render] Texture for plane 0: 1280x720
[libmpv_render] Texture for plane 1: 640x360
[libmpv_render] Texture for plane 2: 640x360
[libmpv_render] Testing FBO format rgba16f
[libmpv_render] Using FBO format rgba16f.
[libmpv_render] No advanced processing required. Enabling dumb mode.
[cplayer] first video frame after restart shown
[cplayer] audio ready
[cplayer] starting audio playback
[libmpv_render] Window size: 2560x1440 (Borders: l=0 t=0 r=0 b=0)
[libmpv_render] Video source: 1280x720 (1:1)
[libmpv_render] Video display: (0, 0) 1280x720 -> (0, 0) 2560x1440
[libmpv_render] Video scale: 2.000000/2.000000
[libmpv_render] OSD borders: l=0 t=0 r=0 b=0
[libmpv_render] Video borders: l=0 t=0 r=0 b=0
[cplayer] playback restart complete @ 0.000000, audio=playing, video=playing
[cplayer] Set property: shared-script-properties -> 1
[statusline] AV: 00:00:00 / 00:00:15 (0%) A-V: -0.000
[cplayer] Set property: shared-script-properties -> 1
[statusline] AV: 00:00:13 / 00:00:15 (90%) A-V:  0.000
[lavf] EOF reached.
[statusline] AV: 00:00:14 / 00:00:15 (98%) A-V:  0.000
[af] filter input EOF
[af] filter output EOF
[cplayer] audio filter EOF
[cplayer] audio draining
[cplayer] audio EOF reached
[statusline] AV: 00:00:14 / 00:00:15 (99%) A-V:  0.000
[vf] filter input EOF
[vf] filter output EOF
[statusline] AV: 00:00:14 / 00:00:15 (99%) A-V:  0.000
[cplayer] EOF code: 1  
[cplayer] finished playback, success (reason 0)
[cplayer] 
[cplayer] Exiting... (End of file)
[cplayer] Set property: shared-script-properties -> 1
[cplayer] draining left over audio
```

birros:
I pushed a refactor, which allows using the same pixel format attributes that we both have with mpv. But I don't see any difference.

I put here the most notable differences I could find between our two setups from the mpv logs:

| libmpv_render  | MacBook Intel (x86)       | MacBook Apple Silicon (arm64) |
| -------------- | ------------------------- | ----------------------------- |
| `GL_VERSION`   | 4.1 INTEL-20.2.50         | 4.1 Metal - 76.3              |
| `GL_VENDOR`    | Intel Inc.                | Apple                         |
| `GL_RENDERER`  | Intel(R) UHD Graphics 617 | Apple M1                      |

alexmercerind:
Sorry I'm little occupied, I had some early look at the PR today.

As far as I can tell, we are still copying the video frames into an intermediate pixel buffer (CPU RAM). This will quite certainly cause redundant CPU usage & drain battery. It's same as software rendering, only difference is that we are still involving GPU on both libmpv & Flutter sides (while still copying frame into CPU RAM in the middle).

This implementation is somewhat like:

libmpv (OpenGL) ---> Pixel Buffer ---> Flutter (METAL).

Since, there's no way (psst... there is ANGLE for exact purpose, translating GLES 2.0 to *) to share a rendered surface/output/frame/texture (whatever) between METAL & OpenGL, this implementation is copying the frame into an intermediate pixel buffer, which then is read by Flutter & sent back to METAL for rendering.

[Flutter recently removed support for OpenGL rendering & OpenGL external texture completely](https://github.com/flutter/engine/pull/37448) i.e. [even Flutter doesn't use OpenGL on Intel Apple machines anymore](https://github.com/flutter/flutter/issues/108304) (does it fallback to CPU based software rendering like on Windows 7?).

There is METAL based external texture API, which we should aim to use for performant zero-copy GPU based rendering: 

- https://github.com/flutter/engine/pull/24523
  - [Implementation PR]
- https://github.com/flutter/engine/pull/37703
  - [AFAIK: They renamed few classes etc. since METAL is only backend now; no OpenGL]

alexmercerind:
As far as my knowledge goes, I this simple software rendering solution will actually perform better than jumping across OpenGL/pixel-buffer/METAL for now. We can use `MPV_RENDER_API_TYPE_SW` rendering to render directly into a pixel buffer.

Until we figure out METAL etc., we can possibly keep supporting it.

I'm open to any ideas or feedback etc. I'm as clueless as everyone about macOS.

birros:
I just realised my mistake, sorry...

The Flutter doc is really bad, even non-existent on this part, I was completely missing `FlutterMetalExternalTexture`.

That doesn't excuse my lack of understanding of how it all fits together, now I understand, my implementation does two GPU renders (Flutter + mpv), passing the data through RAM (`copyPixelBuffer`). To get the most out of the GPU performance of the machine, we need to skip the RAM data flow by rendering directly into a Flutter-ready texture without any intermediary, in this case a Metal texture since OpenGL is no longer supported.

Ok. I'll have to dig on my side to find a solution!

In the meantime, I'll try to modify the code to render mpv directly in the pixel buffer as suggested (`MPV_RENDER_API_TYPE_SW`). Hopefully the performance will improve. Even if as it is, I find the rendering perfectly usable for my needs ^^

But I understand that this is not the ideal way to do it...

alexmercerind:
Yeah! 😄 I also didn't understand this initially. I'm also quite new to this.

One exciting thing I found few days ago is this:

https://gist.github.com/daramkun/456d81800ef076a23caed52f3dd7f808

Notice how GPU to GPU copying is almost 10x faster (& even faster if more capable GPU is taken as reference; _around 25x_) than CPU to CPU copy. While this is just the copy operation which causes that CPU load, there's actual decoding & rendering process which GPU definitely does better. Certainly, CPU isn't good at doing _all the things_.

birros:
Some news :

1. I can make software rendering directly into the `pixelBuffer`, without too much difficulty, I need to clean up the code a bit before commit
2. 🎉 I can render libmpv in a Metal texture thanks to ANGLE! This was very difficult! I have a question about this, I'm using `eglCreateWindowSurface` from an off screen `CALayer` as `EGLSurface`, will this cause a `GPU <-> RAM` copy as `pixelBuffer` or not?
3. I've searched a lot (too much) of time to find out how to access Flutter's Metal texture, with no success, I feel like there is no linux equivalent API yet to do without `copyPixelBuffer`. Does anyone have any ideas? In the meantime, I'll commit what I've done, although the logic still uses `copyPixelBuffer` due to the lack of direct access to the Metal texture. This should fix the Apple Silicon problem, as it will remove the dependency on the host system's OpenGL implementation

alexmercerind:
> 2. :tada: I can render libmpv to a Metal texture thanks to ANGLE! This was very difficult! I have a question about this, I'm using ~`eglCreateWindowSurface` from an off screen `CALayer`~ `eglCreatePbufferFromClientBuffer` from `iosurface` as `EGLSurface`, will this cause a `GPU <-> RAM` copy as `pixelBuffer` or not?

Indeed. After reading few threads online, it seems `ÌOSurface` is the correct way.

On Windows, we pass shared `HANDLE` of the `ID3D11Texture2D` texture in `eglCreatePbufferFromClientBuffer`.
Infact, [Flutter itself uses ANGLE on Windows](https://github.com/flutter/engine/blob/90815e5f75dee1c20ce6d61d5161157513d90b84/shell/platform/windows/angle_surface_manager.cc#L299). Does Skia have native METAL backend (or Skia internally uses ANGLE for macOS)?

<hr>

> 3. I've searched a lot (too much) of time to find out how to access Flutter's Metal texture, with no success, I feel like there is no linux equivalent API yet to do without `copyPixelBuffer`. Does anyone have any ideas? In the meantime, I'll push what I've done, although the logic still uses `copyPixelBuffer` due to the lack of direct access to the Metal texture. This should fix the Apple Silicon problem, as it will remove the dependency on the host system's OpenGL implementation.

We want to get rid of pixel buffers completely. We only need to render using GLES/EGL API provided by ANGLE (in this used by libmpv). On Flutter side, it can already access METAL/IOSurface (?) internally. Drawing using GLES/EGL (libmpv) & reading using METAL/IOSurface (Flutter), where ANGLE is a bridge. A good example can be Flutter itself.

I don't know much at the moment unfortunately.

<hr>

> 4. Bad news: The GPU is much (too) stressed when ANGLE is used instead of the host OpenGL implementation... This corresponds to the use of intermediate structures used in RAM. One clue: mpv selects `videotoolbox-copy` rather than `videotoolbox` with ANGLE, which may explain the excessive consumption…
>    ```
>    log: libmpv_render/videotoolbox error need >= OpenGL 3.0 for core rectangle texture support
>    ```

Windows also falls back to `*-copy` hwdec, still performance is really good (somewhat better than Google Chrome).

<hr>

Off-Topic: This version of ANGLE also seems to support iOS: https://github.com/kakashidinho/metalangle. Let's not worry.

alexmercerind:
I fixed #40, resolved few other things & will polish few more. I find package(s) quite stable on Windows & GNU/Linux.

I feel I should release the current work on pub.dev providing support for the two platforms initially. I will do it later today.

Does that sound right? Any trouble, crashes you guys are facing on Windows or GNU/Linux?

<hr>

I have sent you invitation to this repository @birros, please check your linked e-mail.
You should push your recent work to this branch & I will make sure to go on from there.

Thanks!

JulienDev:
Hello guys, you are both making an awesome job.
Just to let you know, I've tested your current branch @birros on my Macbook Pro (M1 Pro) and the video is blinking as you can see in this video : https://youtu.be/Q7Re8EDXz-M

birros:
As the title of the commit indicates, I've added support for software rendering, so it's not as powerful as OpenGL rendering and doesn't support HDR.

I've also added a basic threading, only for rendering, I'll see if I can improve it.

---

@alexmercerind:
> Indeed. After reading few threads online, it seems IOSurface is the correct way.

> Windows also falls back to *-copy hwdec, still performance is really good (somewhat better than Google Chrome).

Cool! That means I'm on the right way! However, according to my first tests `ANGLE + Metal` is less efficient than `Apple OpenGL + copyPixelBuffer`... If I manage to add Metal support (`TextureMetal`) I think it would be appropriate to keep `TextureGL` and let the choice of which implementation to choose.

---

@JulienDev:

> Just to let you know, I've tested your current branch @birros on my Macbook Pro (M1 Pro) and the video is blinking as you can see in this video : https://youtu.be/Q7Re8EDXz-M

You seem to have the same problem as @damywise. Can you try what I just pushed.
Don't forget to do a `cd macos && pod install` first.

JulienDev:
Congrats! it's way better with your last commit @birros, it's still blinking a bit but not as in my previous video. 
Also, you can notice that the UI is freezing before the first frame is displayed, that's not the case on Windows.
New video: https://youtu.be/I0PzU3NDdvk

alexmercerind:
> Congrats! it's way better with your last commit @birros, it's still blinking a bit but not as in my previous video. Also, you can notice that the UI is freezing before the first frame is displayed, that's not the case on Windows. New video: https://youtu.be/I0PzU3NDdvk

None of this is final!

Th idea is to have a performant H/W accelerated implementation done first which utilizes METAL. We can deal with synchronization issues later. [There has been a ton of flickering in initial days of Windows implementation, which I slowly figured out (see those 43 comments & ton of linked PRs)](https://github.com/alexmercerind/media_kit/issues/10).

TL;DR

This implementation is not "correct" or performant enough (at all) as we need to play 4K/8K videos without draining battery. Spending time to remove this flickering isn't fruitful because this implementation will be changed aggressively anyway to use H/W APIs. Then, we might face different challenges at that point or flickering may be resolved automatically at that new state of work.

<hr>

Off-Topic:

`MPV_RENDER_API_TYPE_SW` will never "flicker" like that (no matter what, even if multi-threaded). But, that isn't the optimal performant solution at all. If we are only targeting software rendering in this PR (while we hopefully figure out ANGLE/METAL in subsequent PR), we don't even need a single trace of OpenGL or METAL to render video output.

JulienDev:
Yes, I'm not complaining at all don't worry and I know it's a work-in-progress implementation. I thought it was good to give some feedback. Let me know if you want me to test something, I have two Macbook Pro (one M1 pro and an older one using Intel).

birros:
> This implementation is not "correct" or performant enough (at all) as required to play 4K/8K videos without draining battery. Spending time to remove this flickering isn't fruitful because this implementation will be changed aggressively anyway to use H/W APIs. Then, we might face different challenges at that point or flickering may be resolved automatically at that new state of work.

II agree with this, you can merge as is if you want. That way, I can create a new pulll request dedicated to improving the existing one, hoping to bring Metal support.

> If we are only targeting software rendering in this PR (while we hopefully figure out ANGLE/METAL in subsequent PR), we don't even need a single trace of OpenGL or METAL to render video output.

We can keep the OpenGL code and set `enableHardwareAcceleration` to `false` by default for macOS or remove it otherwise. Whatever you like.

birros:
@JulienDev:
> it's still blinking a bit but not as in my previous video.

I can finally reproduce this behaviour correctly, it comes from mpv's render time. When mpv renders, it clears the framebuffer before rendering the video frame, if the render time is a bit long and Flutter uses the buffer during this time, it is completely black.

```swift
let black = NSColor(red: 0.0, green: 0.0, blue: 0.0, alpha: 1.0)
OpenGLHelpers.clearScene(context, frameBuffer!, black)
glFlush()

usleep(20000) // wait 20 milliseconds

let red = NSColor(red: 1.0, green: 0.0, blue: 0.0, alpha: 1.0)
OpenGLHelpers.clearScene(context, frameBuffer!, red)
glFlush()
```

> Also, you can notice that the UI is freezing before the first frame is displayed, that's not the case on Windows.

It seems that initializing mpv on your mac takes a long time. Do you have the same problem when using media_kit in audio only mode or only when displaying a video?

damywise:
Performance since the last 3 commits has been incredible! (I've also merged the native_event loop commits on my end, which contributed A LOT).
1. Nothing crashes on any scenario.
2. UI is not lagging/hanging anymore when videos are playing (except when switching tabs).
3. Videos aren't lagging anymore!
4. Black screen blinking is still there but so much less (quite noticeable on stress test but otherwise works almost perfectly on single player)

birros:
@damywise:
> 5. Black screen blinking is still there but so much less (quite noticeable on stress test but otherwise works almost perfectly on single player)

I'm on it, tomorrow I'll try to implement triple buffering, with only double buffering I still have some black frames.

alexmercerind:
@damywise 

> Overall more performant than on Windows (except for the blinking part)

I'm curious.

And how does the CPU usage compare? I'm tired of explaining. This is S/W rendering, it is in no way even close.

I'm pretty sure you're referring "performance" as "less skipped frames" in stress test. But that's coming at cost of excessive CPU usage & reduced battery life of the machine. It's because all videos are rendered on separate thread while Flutter is rendered on separate thread on Windows.
It makes the UI whole lot smoother (& not drop any frames), while videos may drop few frames depending upon system. Similar to Google Chrome, where if too many videos exist, HTML document stays interactive while videos' aren't that prioritized.

On the other hand, there are other parameters like Apple Silicon being better than anything your Windows is running on.

> I've also merged the native_event loop commits on my end, which contributed A LOT.

Yeah! It's because [I disabled event handling from test application in those commits](https://github.com/alexmercerind/media_kit/blob/e898af86b19b5342bd1eaa9676923464820eaaf5/media_kit_test/lib/main.dart#L936). package:media_kit_native_event_loop isn't even supported on macOS yet, if event handling was still enabled, you would have experienced a direct freeze on stress test screen (try passing events: true).

> I'll try to implement triple buffering, with only double buffering I still have some black frames.

I'm just confused. Why do we need OpenGL, frame buffers or anything in `MPV_RENDER_API_TYPE_SW`?
There will not be any blinking inherently.

birros:
@alexmercerind:
> I'm just confused. Why do we need OpenGL, frame buffers or anything in `MPV_RENDER_API_TYPE_SW`? There will not be any blinking inherently.

We misunderstood, this pull request contains two types of renderings:

- SW: Same as you did for windows & linux, mpv renders directly into a buffer backed by a `CVPixelBuffer`, the GPU is never used in this part
- GL: Here the GPU is used, however the texture rendered by mpv is stored in RAM in a `CVPixelBuffer`, Apple has set up a system that allows to bind an OpenGL texture to this `CVPixelBuffer`. The architecture is `OpenGL + pixelBuffer`.

In both cases, the `CVPixelBuffer` is interoperable with Metal, which allows Flutter to use this `CVPixelBuffer` out of box.

I myself am amazed at the performance, even using an intermediate buffer in RAM, I get better performance than with `ANGLE + Metal`. Simply because `hwdec` performs better when not `*-copy`, which is the case with OpenGL and not ANGLE, because the latter uses OpenGL ES which seems to be compatible only with those in `*-copy`.

damywise:
> how does the CPU usage compare? 

On stress test:

MacOS usage:
image
Windows usage:
image

I misunderstood. I thought that media_kit is gonna prioritize performance over battery even though it's the other way, which is what I assume mpv player does.

> I'm pretty sure you're referring "performance" as "less skipped frames" in stress test.

Yeah that's what I meant, and videos are laggy even with just two players. Sorry if I have used incorrect word! (and/or wasn't specific enough)

> try passing events: true

Indeed that's what makes the UI lags less. I didn't read any of the commits, sorry about that!

alexmercerind:
> We misunderstood, this pull request contains two types of renderings.

Oh, I apologise.

I thought we completely switched to `MPV_RENDER_API_TYPE_SW` . I think you optimized the existing code path.
_i.e_ GL (libmpv) --> Pixel Buffer --> METAL (Flutter)


> Simply because `hwdec` performs better when not `*-copy`, which is the case with OpenGL and not ANGLE, because the latter uses OpenGL ES which seems to be compatible only with those in `*-copy`.

That's completely true. AFAIK it's impossible to achieve non `*-copy` `hwdec` with ANGLE since it's GL ES 2.0. [In this issue comment, @wm4 has mentioned, "There is no way of getting direct hardware decoding with this, or anything better than GLES 2.0."](https://github.com/mpv-player/mpv/issues/7769#issuecomment-635867973).

None the less, we are involving a single copy to RAM on both platforms (Windows & Linux). Either libmpv internally does the RAM/pixel-buffer copy (by switching to `*-copy` hwdec) or in this case we ourselves copy the frame into a pixel buffer for dealing with METAL interop.

Even on Windows, the default `enableHardwareAcceleration` falls back to `dxva2-copy` by default, likely due to same EGL 2.0 reason.

From my personal experience (I'm referring to mpv application), `*-copy` callback actually performs better than direct `hwdec` on my machine. I get frame drops in 4K 60 FPS with `d3d11va`, where as `d3d11va-copy` plays smooth (same with `dxva2` & `dxva2-copy`). So, maybe having CPU involved along-side GPU for that single copy increases the overall performance?

> Ideally, there would be a non `*-copy` `hwdec` with METAL, but I don't think that's possible at the moment, even with ANGLE, until mpv renders to METAL itself.

This is what I was planning for. However, it seems this is also fine as you guys are finding it performant.

One irony is that GNU/Linux implementation doesn't perform any CPU RAM copy, still it feels the slowest.
Still, "theoretically" it is most efficient.

birros:
@damywise or @JulienDev: Can you test what I have pushed.

Normally, flickering should be reduced (or even eliminated) and threading improved

alexmercerind:
Also: @Quackdoc

damywise:
Alright everything seems good!
1. No flicker/blinking/lag
2. UI doesn't lag, freeze, or anything during any occasions!

Very usable

JulienDev:
I can confirm, it's working like a charm, that's huge!

ProgramerMartin:
您好，macos分支的功能简直太棒了，我在测试中也遇到了同样的问题:

birros:
Some news:

Still alive! I finally built **libmpv** and its dependencies from scratch, locally, only for `x64` for now. Next steps: `arm64`, then **GitHub actions**.

birros:
libmpv build part done!

I set up the build from scratch of libmpv for macOS, which is complicated to do "easily", because of the crappy developer experience of this system.

I made sure to respect these rules:
- **"reproducible"**: i.e. to not depend on the libs of the host system and therefore easily deployable on a CI
- **future proof**: for this, I cross compile to both `amd64` & `arm64` architectures, independently of the host architecture
- **commercial use ready**: I chose the right options to build a version of libmpv and its dependencies to be compatible with commercial use (@alexmercerind to check)

I took some liberties with the IINA libs:
- I replaced **GnuTLS** by **LibreSSL**, the former being a pain to build and the latter being maintained by the OpenBSD team (a guarantee of quality!)
- I deactivated a lot of dependencies that I consider superfluous for the use made in **media_kit** (for example x11, screenshot...)
- I didn't add two dependencies that may be important for some, but that I consider optional (for now):
    - libbluray: for reading and parsing Blu-ray discs
    - libarchive: for reading various archive formats, including tar and zip

I tested on `amd64`, it works, with https and subtitle support like the IINA libraries. Can anyone test on `arm64` please? (@damywise, @JulienDev or @Quackdoc)

---

Repo : [birros/libmpv-build](https://github.com/birros/libmpv-build)

damywise:
It works great!

alexmercerind:
The work you've done for making this possible is phenomenal. 👏👏👏

I had to really struggle for getting Windows cross-compile builds running on GitHub actions.
I don't even want to imagine how hard it's for macOS & not to forget, two separate architectures.

<hr>

I would really like to move whole package:media_kit & all the repositories etc. to a separate GitHub organization sooner or later. The scale of things is going to get really large (in some ways, it already is) & we all have big plans.

birros:
LGTM, ready to merge!

@alexmercerind I'll let you handle the conflicts, if you want to modify the doc 🚀

birros:
So, in order to merge with the main branch:
- I added native event loop support, it works as excepted. macOS gave me a hard time, it causes a duplicate download of the mpv headers on install, as I can't force media_kit_video to use headers outside its location. This isn't critical, but it's good to know for future updates, there will be two Makefiles to edit
- Unfortunately, the console is flooded with a lot of warnings which I can't mute, especially when starting in dev mode or building release
- Also, I noticed a degradation of perfs in the stress test tab, which is normal since we move from 4 videos in the previous test to 20 now. But when I go back to 4, there is no difference with the previous perfs, so for me it's ok

I list here all the warnings I encounter every time and which are not critical from my point of view, just confusing for newcomers (maybe to be placed somewhere in the doc):

```
# 1 "<command line>" 1
 ^
<command line>:20:9: warning: 'POD_CONFIGURATION_DEBUG' macro redefined
#define POD_CONFIGURATION_DEBUG 1 DEBUG=1 
        ^
#define POD_CONFIGURATION_DEBUG 1
        ^
```

```
#import "Headers/media_kit_video-Swift.h"
        ^
/path/to/media_kit/media_kit_test/build/macos/Build/Products/Debug/media_kit_video/media_kit_video.framework/Headers/media_kit_video-Swift.h:270:31: warning: 'objc_ownership' only applies to Objective-C object or block pointer types; type here is 'CVPixelBufferRef' (aka 'struct __CVBuffer *')
- (CVPixelBufferRef _Nullable __unsafe_unretained)copyPixelBuffer SWIFT_WARN_UNUSED_RESULT;
```

PR #46, opened 2023-02-25
package:media_kit_native_event_loop

### LINKED ISSUE #40: [Bug] 8 or more Player(s) freeze the application

alexmercerind:
I don't have much time to document how it'd work, implementation should be self-explanatory.

A drop-in solution to resolve the issue. Simply adding the package to `pubspec.yaml` will allow customers to use more than 8 instances of `Player` simultaneously without any issues. [`package:media_kit`](https://github.com/alexmercerind/media_kit/tree/main/media_kit) will switch to this implementation automatically if the shared library generated by this target is found. Otherwise, it will fallback to default [`Isolate` based event handling](https://github.com/alexmercerind/media_kit/blob/main/media_kit/lib/src/libmpv/core/initializer.dart).

```yaml
dependencies:
  media_kit_native_event_loop: ^1.0.0
```

Closes #40.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "media_kit_native_event_loop/src/dart_api"]
-	path = media_kit_native_event_loop/src/dart_api
-	url = https://github.com/alexmercerind/dart_api.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "media_kit_native_event_loop/src/dart_api"]
+	path = media_kit_native_event_loop/src/dart_api
+	url = https://github.com/alexmercerind/dart_api.git
--- a/README.md
+++ b/README.md
@@ -177,6 +177,10 @@ class MyScreenState extends State<MyScreen> {
 }
 ```
 
+### Performance
+
+Although [package:media_kit](https://github.com/alexmercerind/media_kit) is already fairly performant, by following ways:
+
 **Note**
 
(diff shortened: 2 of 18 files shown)


alexmercerind:
This also adds a new API to disable events completely. It can further squeeze the performance if you don't rely on events.

> **Note**
> 
> - You can disable event callbacks for a `Player` & save yourself few CPU cycles.
> - By default, `events` is `true` i.e. event streams & states are updated.
> 
> ```dart
> final player = Player(
>   configuration: PlayerConfiguration(
>     events: false,                              // default: true
>   ),
> );
> ```

alexmercerind:
And... yeah obviously we're no longer limited to just 7 simultaneous `Player`s.
The test application is now rendering 20 HD videos simultaneously on my integrated GPU.

alexmercerind:
@DomingoMG, you should try adding this to your project.

You shouldn't see that 8 `Player`s lag/deadlock anymore. I tried to reach you, maybe you aren't that active on Discord.

Thanks!

PR #60, opened 2023-03-11
feature: change video aspect ratio.

cst1412:
- set video aspect ratio override
- get current video aspect ratio override
- clear current video aspect ratio override

I tested if this is possible via the AspectRatio widget but this did not work for me so i implemented this feature :).

-> MPV docs: https://mpv.io/manual/master/#options-video-aspect-override

### DIFF
--- a/media_kit/lib/src/libmpv/player.dart
+++ b/media_kit/lib/src/libmpv/player.dart
@@ -353,6 +353,38 @@ class Player extends PlatformPlayer {
     calloc.free(args);
   }
 
+  // overrides the current video tracks aspect ratio
+  // https://mpv.io/manual/master/#options-video-aspect-override
+  @override
+  FutureOr<void> setVideoAspectRatioOverride(String aspectRatio) {
+    return setProperty("video-aspect-override", aspectRatio);
+  }
+
+  // gets the current video tracks aspect ratio override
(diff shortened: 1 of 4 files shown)


alexmercerind (changes requested):
Is there no way to work within the `Video` widget itself?

Expose a simple `aspectRatio` optional argument (`null` by default) in `Video` constructor.

You can try using the [`AspectRatio`](https://api.flutter.dev/flutter/widgets/AspectRatio-class.html) widget from Flutter itself. You can also [get current video dimensions as a `ValueNotifier<Rect>`](https://github.com/alexmercerind/media_kit/blob/7cf5d63e74d57fe858083b7fee52e36984a7e1c3/media_kit_video/lib/src/video.dart#L114) updated by native implementation.

A basic thing like aspect ratio should be handled on Flutter side itself using widgets & basic handling etc. This will make us less subjected to platform specific behaviour (libmpv) & prevent any "odd"-ish surface API as such in `Player`.

<hr>


Thanks a lot for the pull request, I really appreciate it.
I really want to ensure the quality of the code everyone uses & depends upon. Hope you understand.

I'm requesting changes for now.

cst1412:
If this is possible you are probably right. I will try when i have time. Thanks for your suggestion :)

alexmercerind:
See #76.

PR #61, opened 2023-03-15
rename media_kit_libs_macos to to media_kit_libs_macos_video

birros:
Small refactor needed to avoid future changes (libs for audio only).


alexmercerind (approved):
Thanks for addressing the feedback!

LGTM.

PR #70, opened 2023-03-19
add iOS support

birros:
🎉 Add iOS support

## What works in this pull request

- [x] Software rendering, directly in a pixel buffer (RAM)
- [x] "Partial" hardware rendering, using OpenGLES + pixelBuffer, as macOS does

## What remains to be done in this pull request

- [x] Try to solve HDR color
- [x] Try to share code between macOS & iOS, currently a lot of code is duplicated
- [x] Try to fix frameBuffer error, ~~which appears in the logs but does not affect the playback~~
- [x] Add `native_event_loop` support
- [x] Add CI build
- [x] Update `README.md`


birros:
Ready to merge!

I deduplicated some code to share it between iOS and macOS, and I also fixed a weird bug when creating the frameBuffer by adding a retry to it.

PR #72, opened 2023-03-21
add CI build for macOS

birros:
Add CI universal build for macOS (amd64 & arm64).
I tested the artifact, it works.

I also added a condition on the release creation so that it only happens on the main branch: [.github/workflows/ci.yml#L82](https://github.com/birros/media_kit/blob/3a0817296155635dab491d4594fac9c89aa0df35/.github/workflows/ci.yml#L82)

@alexmercerind If it's ok with you, we can merge.


alexmercerind (approved):
Thanks a lot!

PR #76, opened 2023-03-23
feat: aspect ratio in Video Widget

alexmercerind:
Adds optional `aspectRatio` argument to `Video` widget.

```dart
class _MyVideoState extends State<MyVideo> {
  @override
  Widget build(BuildContext context) {
    return Video(
      controller: controller,
      // 🎉
      aspectRatio: 1.0,
    );
  }
}
```

<hr></hr>

Implements feature requested in #60.


wengxianxun:
The first setting of aspectRatio is valid, but when I change a value and refresh it later, it becomes invalid

wengxianxun:
how to call update method？

PR #81, opened 2023-03-24
fix: broken hot restart when using native_event_loop (macOS)

### LINKED ISSUE #74: [Bug Report] native_event_loop: Hot restart interrupts the listening of events
This is a minor bug, as it does not affect an app in production, only in development.

When the `native_event_loop` is used, if we hot restart the app in dev mode the mpv events are no longer listened by the `Player`.

birros:
Fix #74

### DIFF
--- a/media_kit_native_event_loop/src/media_kit_native_event_loop.cc
+++ b/media_kit_native_event_loop/src/media_kit_native_event_loop.cc
@@ -16,12 +16,10 @@ MediaKitEventLoopHandler& MediaKitEventLoopHandler::GetInstance() {
 void MediaKitEventLoopHandler::Register(int64_t handle,
                                         void* post_c_object,
                                         int64_t send_port) {
-  auto post_c_object_ =
-    reinterpret_cast<bool (*)(Dart_Port, Dart_CObject*)>(post_c_object);
+  std::thread([&, handle, post_c_object, send_port]() {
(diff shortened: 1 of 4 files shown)


birros:
I only removed the global `send_port`, but I can do the same for `post_c_object` if you want?

media_kit_native_event_loop/src/media_kit_native_event_loop.cc
```diff
@@ -22,12 +22,9 @@ void MediaKitEventLoopHandler::Register(int64_t handle,
     post_c_object_ =
         reinterpret_cast<bool (*)(Dart_Port, Dart_CObject*)>(post_c_object);
   }
-  if (send_port_ == -1) {
-    send_port_ = send_port;
-  }
 
   auto context = reinterpret_cast<mpv_handle*>(handle);
-  std::thread([this, context]() {
+  std::thread([this, context, send_port]() {
```
alexmercerind, line 27:
```suggestion
  std::thread([=]() {
```

Should work AFAIK.

alexmercerind:
> I only removed the global `send_port`, but I can do the same for `post_c_object` if you want?

Sure, let's remove them all. And this comment too:

https://github.com/alexmercerind/media_kit/blob/c5b85d1e679b733ff193c41ada52546034133a17/media_kit_native_event_loop/src/media_kit_native_event_loop.cc#L19-L20

🙏

PR #85, opened 2023-03-26
feat: add VideoController.setSize

### LINKED ISSUE #84: [Enhancement] Possibility to resize the rect of `VideoController`
From the documentation:

```dart
/// 1. You can limit size of the video output by specifying `width` & `height`.
///    By default, both `height` & `width` are `null` i.e. output is based on video's resolution.
```

Currently, I initialize the video with `640x360`, but I find the need to use a smaller resolution when the video is resized (192x144). Would it be possible to change the height and width of the `VideoController`?

I tried to create different video controllers with different resolutions, but it seems that the second controller isn't created if one controller is already initialized. It also crashes when the second controller is initialized. Is that by design?

birros:
Fix #84

Drawbacks: the same as when you set the size with `VideoController.create`, when the video is paused, if you upscale the widget, the texture is not upscaled.

## Platforms

- [x] macOS
- [x] iOS
- [x] Linux
- [x] Windows
- [x] Android

## Usage / Example

```dart
controller?.setSize(width: 1920, height: 1080);
```


birros:
Ideally, a `maxWidth` would be more relevant, as this would avoid unnecessarily upscaling a low definition video.

alexmercerind:
I can't push to your branch. May you do something? I'd need it for #113 too.

alexmercerind:
Nevermind.

alexmercerind:
I changed API a bit (if that's fine, made changes for macOS too).

alexmercerind (approved):
LGTM from my end.

I've added support for Windows, GNU/Linux & Android. Android implementation is based on `--android-surface-size` from libmpv, doesn't really seem to do anything. I'm fine since this API is already an overkill.

birros (commented):
LGTM!

birros:
I am on it!

PR #90, opened 2023-03-28
Major Changes

alexmercerind:
1. Now `Media` can be opened directly.

```dart
await player.open(Media('file:///C:/Users/Hitesh/Music/Sample.mp3'));
await player.open(Media('file:///C:/Users/Hitesh/Video/Sample.mkv'));
await player.open(
  Playlist(
    [
      Media('file:///C:/Users/Hitesh/Video/Sample.mkv'),
      Media('https://www.example.com/sample.mp4'),
      Media('rtsp://www.example.com/live'),
    ],
  ),
);
```

2. Improve `Player`'s internal playlist & state handling.
3. Draw first frame on `Player.open` before `Player.play`.
4. Bump required Flutter version to 3.7.0 in package:media_kit_video.
5. Utilize the unregister texture release callback on Windows. Further improves overall stability.
6. Updated CHANGELOG.
7. [Test application now has test screens in separate individual files](https://github.com/alexmercerind/media_kit/tree/code-cleanup-and-restructure/media_kit_test/lib/tests).
8. New test: [06.paint_first_frame.dart](https://github.com/alexmercerind/media_kit/blob/code-cleanup-and-restructure/media_kit_test/lib/tests/06.paint_first_frame.dart).

PR #94, opened 2023-03-29
fix: playing state/stream updates & compatibility with libmpv 0.28.x

alexmercerind:


alexmercerind:
This was one regression in last pull-request #90.

Having compatibility with older libmpv versions is a concern for GNU/Linux, you never know what people may install...

PR #107, opened 2023-04-04
fix: double play pause

zezo357:
changed the calls of 

- play to 'set pause no'
- pause to 'set pause yes'
 
fixes:
```
player.open(Media('asset://assets/video_$i.mp4'));
player.play();
```
pausing the player when it should play


zezo357:
Hello,

This pull request was marked as a draft, even though it's ready. Should I take any action?

alexmercerind:
Thanks for PR! I need to review & check across few libmpv versions.

media_kit_test/lib/tests/01.single_player_single_video.dart
```diff
@@ -64,6 +64,7 @@ class _SinglePlayerSingleVideoScreenState
             ),
             onTap: () {
               player.open(Media('asset://assets/video_$i.mp4'));
+              player.play();
```
alexmercerind, line 67: Why?

media_kit_test/lib/common/widgets.dart
```diff
@@ -183,7 +183,7 @@ class _SeekBarState extends State<SeekBar> {
       children: [
         const SizedBox(width: 48.0),
         IconButton(
-          onPressed: widget.player.playOrPause,
+          onPressed: playing ? widget.player.pause : widget.player.play,
```
alexmercerind, line 186: Why?

alexmercerind:
I made few fixes other than this after discovering some regressions in [Harmonoid](https://harmonoid.com). Ultimately, this issue automatically got resolved. Still... I replaced calls to remove any race between `pause` event update & actually changing `pause` property.

media_kit_test/lib/common/widgets.dart
zezo357: To avoid confusion while testing, since using playorpause can do the same function

media_kit_test/lib/tests/01.single_player_single_video.dart
alexmercerind: It's not right to make redundant changes to the test application. This test application is to test the rendering etc.
The playback, playlist, event streams etc. should be tested as part of `media_kit/*` or `media_kit/tests/*`.

alexmercerind:
Hi @zezo357,

May you please test the current [tip of the `main` branch](https://github.com/alexmercerind/media_kit/commit/cb8163af90821bf4259bb65a2c0354a279074cf0)? I can't seem to reproduce it + it fixes one another issue.

media_kit_test/lib/tests/01.single_player_single_video.dart
zezo357: > It's not right to make redundant changes to the test application. This test application is to test the rendering etc.
> The playback, playlist, event streams etc. should be tested as part of `media_kit/*` or `media_kit/tests/*`.

Ok, so you want me to remove it, and add it to media_kit test right?

zezo357:
> Hi @zezo357,
> 
> May you please test the current [tip of the `main` branch](https://github.com/alexmercerind/media_kit/commit/cb8163af90821bf4259bb65a2c0354a279074cf0)? I can't seem to reproduce it + it fixes one another issue.
> I'll publish an update for it.

Ok will test and let you know.

zezo357:
> Hi @zezo357,
> 
> May you please test the current [tip of the `main` branch](https://github.com/alexmercerind/media_kit/commit/cb8163af90821bf4259bb65a2c0354a279074cf0)? I can't seem to reproduce it + it fixes one another issue. I'll publish an update for it.

https://github.com/zezo357/media_kit/tree/test_double_play_main still happens 

if you are playing it works correctly but if you pause and try another file it doesnt start

zezo357:
> You mean:
> 
> 1. Click on a video to play.
> 2. Pause using play/pause button.
> 3. Click on another video to play.
> 4. It doesn't play.
> 
> ???
> 
> I can't reproduce. Also... Please add `await` in those method calls, chances are they can conflict (they are `async`).

making them await fixes the issue (thats on main)

i guess there is no need to my pr then since it can be solved by awaiting?

alexmercerind:
I'll add internal synchronization then.

PR #112, opened 2023-04-06
Fixed glitching slider while seeking in test project.

akaMrNagar:
Solved issues :

1. Slider glitching back and forth while seeking.
2. Disabled the slider to prevent the user from seeking even if the media is ended.


alexmercerind (approved):
LGTM!

I don't really care for test application, it's just for development & debugging. But since many people may see it as an example project showing "best practices", I feel it's good to merge.

PR #122, opened 2023-04-11
fix: assets with special characters

### LINKED ISSUE #121: Does not handle Japanese character
FIrst of all, thank you for this brillant package!
However, I had some problems getting started because this package does not seem to handle some characters (correctly).

I have a file called `う.wav` which cannot be played. I think this is not a problem but there is on error message (or anything) indicating that the name is the problem, the sound just does not play. Renaming the file to `u.wav` resolved the issue

I am converting from `just_audio` to this package (in just audio the name works) and this took me quite some time to figure out, therefore, a note somewhere would be nice (in my opinion).

alexmercerind:
Closes #121.


birros:
I recommend adding a small unit test to cover this case... and the following potential cases...
Even if it takes a bit more time to fix.

alexmercerind:
I really agree for the tests. We seriously need unit tests inside package:media_kit.

There's a lot to cover at the moment. The rendering size is still well tested.

birros:
I added a unit test, I don't know the Dart / Flutter conventions, I did what I thought was right.

birros (approved):

alexmercerind:
Converting to draft for now based on the last comment on issue thread.

media_kit/lib/src/models/media.dart
```diff
@@ -46,58 +47,59 @@ class Media extends Playable {
...
         );
+      } else if (Platform.isAndroid) {
+        asset = path.normalize(AndroidAssetLoader.instance.load(key));
+      } else {
+        throw UnimplementedError(
+          '$_kAssetScheme is not supported on ${Platform.operatingSystem}',
+        );
+      }
+      if (!File(asset).existsSync_()) {
+        throw Exception('Unable to load asset: $asset');
```
alexmercerind, line 100: Now throwing an exception if an asset is not located.

Issue #35, opened 2023-02-02
[Feature Request] Buffering progress


bdlukaa:
An option to know the current buffering progress. This is useful to use in a slider alongside with the current player position

alexmercerind:
I have to see if that's possible. Currently working on other issues.

Quackdoc:
I don't have the time to work on this yet but it's likely possible to make use of the [demuxer-cache-state](https://mpv.io/manual/master/#command-interface-demuxer-cache-state) property for this. it's also possible to set custom cache size using mpv properties. it returns the value as `cache-end` or we can use the dedicated property `demuxer-cache-time`. both are the same value. demuxer-cache-state will simple also return other values as listed in the link.

PR #123, opened 2023-04-12
feat: buffer stream & state

zezo357:
Added functionality to retrieve the last buffered placement in the player and implemented the necessary logic in the media_kit_test to test the changes. Please refer to the attached photo for reference.

Changes Made:

- Added code to retrieve the last buffered placement in the player.
- Implemented logic in the media_kit_test to test the changes.
- Attached a photo for reference.

### Note: tested on windows

image


alexmercerind:
Closes #35.

The thing to note here is that `Player` will not willingly try to decode whole stream in one go.
By default, we are limiting demuxer cache to about 32 MB.

I'm satisfied with implementation since it follows the existing structure.

alexmercerind (approved):
LGTM!

I renamed it to just `buffer` & I think it correctly expresses the API.
All 3 `position`, `duration` & `buffer` streams/states look & work alike.

PR #127, opened 2023-04-13
feat/refactor: entry-point API

alexmercerind:
Adds new API `MediaKit.ensureInitialized`, this allows to:
- Handle libmpv & package:media_kit_native_event_loop discovery + loading in advance.
- Receive some additional initialization arguments etc.
- Expose ability to handle/catch any initialization errors to clients e.g. libmpv shared library not found etc.
- Dispose already running `Player` instances on hot-restart.

Other changes:
- Set `PlayerConfiguration.vo` to `null` by default.
  Windows, GNU/Linux & Android embeddings now set it to `libmpv` when a new `VideoOutput` is created.
- Refactor libmpv `mpv_handle*` initialization into `InitializerIsolate` & `InitializerNativeEventLoop` classes.

### DIFF
--- a/media_kit/lib/media_kit.dart
+++ b/media_kit/lib/media_kit.dart
@@ -7,6 +7,7 @@
 // ignore_for_file: camel_case_types
 
 export 'package:media_kit/src/player.dart';
+export 'package:media_kit/src/media_kit.dart';
 export 'package:media_kit/src/platform_player.dart';
 
 export 'package:media_kit/src/models/media.dart';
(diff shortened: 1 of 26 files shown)


alexmercerind:
Not sure if we should land it or not. Kind of opinionated & breaks public API a bit.

birros (approved):
LGTM!

I repeat what I said yesterday on Discord, it improves the developer experience.
I've just tested it, it kills instances still alive on a hot restart. It's a life changer (during development).
So yes, it's a good idea to merge it, sooner or later there will be issues about it…

PR #130, opened 2023-04-14
feat: add audio libs (macOS & iOS)

birros:
I've bumped `media_kit_libs_(macos|ios)_*` to 1.0.4, to make the "darwin" platform consistent.


birros:
@alexmercerind I'll let you review and merge (and publish), I think publishing at least one new version before v1 would allow me to test the "optional libs" in real conditions.

PR #146, opened 2023-04-19
feat: add httpHeaders configuration

lucinhu:


alexmercerind:
We need a `headers` argument in [`Media` constructor](https://github.com/alexmercerind/media_kit/blob/050e039f2a5395022acafb5477b63f804cfe13cd/media_kit/lib/src/models/media.dart#L38-L44) instead. Different headers may be required for different media sources, having a single time initialization parameter is not good. It is planned; I'm currently busy with other things. We're growing :)

alexmercerind:
[libmpv hooks](https://mpv.io/manual/master/#hooks) is a potential solution for achieving the API-style I recommended above.
A general idea is to set the headers property right before the media starts [; store available headers in a global `HashMap` like this for later access](https://github.com/alexmercerind/media_kit/blob/050e039f2a5395022acafb5477b63f804cfe13cd/media_kit/lib/src/models/media.dart#L16-L17).

For now you can use [`setProperty` API](https://github.com/alexmercerind/media_kit/issues/14#issuecomment-1373205132) as alternative to this approach.

In-case you wish to implement this, we can review from there. Thanks!

lucinhu:
> [libmpv hooks](https://mpv.io/manual/master/#hooks) is a potential solution for achieving the API-style I recommended above. A general idea is to set the headers property right before the media starts [; store available headers in a global `HashMap` like this for later access](https://github.com/alexmercerind/media_kit/blob/050e039f2a5395022acafb5477b63f804cfe13cd/media_kit/lib/src/models/media.dart#L16-L17).
> 
> For now you can use [`setProperty` API](https://github.com/alexmercerind/media_kit/issues/14#issuecomment-1373205132) as alternative to this approach.
> 
> In-case you wish to implement this, we can review from there. Thanks!

thanks!

PR #128, opened 2023-04-13
[WIP] feat: web support

alexmercerind:
Support for Flutter Web.

The last (but not least) platform package:media_kit didn't have support for. Once this pull-request lands:

| Platform  | Video | Audio |
| --------- | ----- | ----- |
| Android   | ✅ | ✅ |
| iOS       | ✅ | ✅ |
| macOS     | ✅ | ✅ |
| Windows   | ✅ | ✅ |
| GNU/Linux | ✅ | ✅ |
| Web       | ✅ | ✅ |

PR #148, opened 2023-04-20
fix/refactor: improve player lifecycle

alexmercerind:
## Changes

- fix: S/W rendering fallback & `--vo=gpu` rendering on Android
- fix: video output in Android emulator (#149)
- fix: now `playlist` state/stream is populated from libmpv event instead of manually updating inside `open`/`add`/`move` etc.
- feat: preserve last rendered video frame upon `Media` playback completion
- feat(windows): `enableHardwareAcceleration` configurability 

## Others

- chore: add button to toggle between H/W & S/W rendering in test application
- chore: test `06.paint_first_frame.dart` with non-zero playlist index

### DIFF
--- a/media_kit_video/lib/src/video_controller_android.dart
+++ b/media_kit_video/lib/src/video_controller_android.dart
@@ -11,6 +11,7 @@ import 'package:ffi/ffi.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter/foundation.dart';
 import 'package:media_kit/media_kit.dart';
+import 'package:synchronized/synchronized.dart';
 
 // ignore_for_file: unused_import, implementation_imports
 // It's absolutely crazy that C/C++ interop in Dart is so much easier & less tedious (possibly more performant as well) than in Java/Kotlin.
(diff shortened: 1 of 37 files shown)


alexmercerind:
Personally speaking, these changes really improve the overall quality & stability.

Now, I aim to add some solid unit tests inside package:media_kit (running on each platform), primarily testing:
- Playback of different kinds of `Media` sources.
- Parsing of various kinds of URI(s) in `Media` constructor.
- State changes & events in `Player` e.g. `playing`, `completed` etc. upon calling various method calls e.g.
  - `open` with `play: true`
  - `open` with `play: false`
  - `playOrPause`
  - etc.
- Misc. state changes & events in `Player` like `bitrate`, `width`, `height`, `volume`, `rate`, `buffering` etc.

I think now is the best time to do it because now logic inside `Player` class is _finalized_. Having these tests will allow us to:
- Handle regressions & account for changes in behavior due to any updates inside package:media_kit.
- Handle regressions & account for changes in behavior due to libmpv updates.
- Help the upcoming web implementation (#128) reach the similar surface API/behavior.

<hr>

For the video rendering part, [media_kit_test](https://github.com/alexmercerind/media_kit/tree/main/media_kit_test) is doing it's duty quite well. All the cases (e.g. single `Player` + single `Video`, single `Player` + multiple `Video`, multiple `Player` + multiple `Video` etc.) are covered & we may add more strict ones (like [`06.paint_first_frame.dart`](https://github.com/alexmercerind/media_kit/blob/main/media_kit_test/lib/tests/06.paint_first_frame.dart)) in future. [`04.tabs_test.dart`](https://github.com/alexmercerind/media_kit/blob/main/media_kit_test/lib/tests/04.tabs_test.dart) is specifically present to look for any race/synchronization issues or access violations etc. Hopefully, a human being can verify rendering/playback of a video better than some integration test. Manually verifying all rendering conditions before every stable release is fine enough & I'd still do it anyway even if we get some integration test in future.

birros:
I totally agree.

> Manually verifying all rendering conditions before every stable release is fine enough & I'd still do it anyway even if we get some integration test in future.

Indeed, I think that automating the testing of the video rendering part is probably "impossible".
However, it's possible to set up a checklist template covering all the cases to be tested manually before a release.
As well as post-release testing, for example to check that the versions of the packages listed in the documentation exist…

media_kit_video/lib/src/video_controller_android.dart
```diff
@@ -74,13 +116,21 @@ class VideoControllerAndroid extends VideoController {
       return _controllers[handle]!;
     }
 
+    // Enforce software rendering in emulators.
+    final bool isEmulator = await _channel.invokeMethod('Utils.IsEmulator');
```
birros, line 120: I use the same logic to force SW rendering for the iOS simulator, however the logic is in `VideoOutput.swift`, maybe I should move the logic to `video_controller_native.dart` when the platform is iOS, it would make the code more consistent.
What do you think?

media_kit_video/lib/src/video_controller_android.dart
```diff
@@ -144,27 +214,10 @@ class VideoControllerAndroid extends VideoController {
...
-    NativeLibrary.ensureInitialized();
-    final mpv = MPV(DynamicLibrary.open(NativeLibrary.path));
-    final name = 'android-surface-size'.toNativeUtf8();
-    final value = '${width}x$height'.toNativeUtf8();
-    mpv.mpv_set_option_string(
-      Pointer.fromAddress(handle),
-      name.cast(),
-      value.cast(),
+  }) {
+    throw UnimplementedError(
```
birros, line 218: I think it should be added to the changelog that a throw has been added to `setSize`, otherwise some people will be surprised ^^

alexmercerind: Sure!

media_kit_video/lib/src/video_controller_android.dart
alexmercerind: I think your implementation is good already (no need for change). I think there's no point in looping back to Dart for this.

It's just that most implementation of Android's video embedding is within Dart (instead of Java/Kotlin) & we need to query this from Dart. A dedicated specialized class was necessary `VideoControllerAndroid` requiring abstraction of `VideoController`, as we know.

alexmercerind:
I added few more tests, primarily for "internal" things. Next, I can move onto testing `Player` class itself (& more meaningful).

https://github.com/alexmercerind/media_kit/blob/da9150a563e855647e6da1d191f107f548a3128e/media_kit/test/src/libmpv/core/initializer_test.dart#L7-L108
https://github.com/alexmercerind/media_kit/blob/da9150a563e855647e6da1d191f107f548a3128e/media_kit/test/src/libmpv/core/native_library_test.dart#L7-L33
https://github.com/alexmercerind/media_kit/blob/da9150a563e855647e6da1d191f107f548a3128e/media_kit/test/src/models/media_test.dart#L7-L146

alexmercerind:
Can we land this? I tested the playlist logic inside my project & everything is working correctly (without any regressions).

<hr>

I will add more tests in future PR. We will need downloading/extracting libmpv shared libraries to an "accessible" location for running tests on Windows/macOS.  For now, these tests are running on GNU/Linux.

https://github.com/alexmercerind/media_kit/blob/da9150a563e855647e6da1d191f107f548a3128e/.github/workflows/ci.yml#L5-L21
https://github.com/alexmercerind/media_kit/blob/da9150a563e855647e6da1d191f107f548a3128e/.github/workflows/ci.yml#L41-L57

birros (approved):
LGTM!

birros:
I tried to use the macOS libs to run the tests, unfortunately impossible to override the `DYLD_LIBRARY_PATH` environment variable, probably because the code is run in a simplified dart VM.

We'll probably must add a script to set hard paths in the dylibs.
Fortunately, we can run cli commands from a test file.

PR #178, opened 2023-05-06
fix(android): shutdown leak

Snowcrash5:
When the `Player::dispose` function is called, the `mpv` instance shutdown is induced via `mpv_command_string` with the command `quit 0`.

On android, `quit 0` does not actually clean some resources (a thread is left active, consuming high CPU). `mpv_destroy` must be called for actual cleaning to happen.

When the player is shut down, a `MPV_EVENT_SHUTDOWN` event is emitted to the event loop, causing it to quit.
In this PR, we simply call `mpv_destroy` on Android devices.

We also wait for `MPV_EVENT_SHUTDOWN` to be emitted before finishing disposing.

I reproduced and tested it out by having an android app display a video and then unload it, over and over again.

It must be noted that after 100+ times, cpu usage becomes high again, and new videos stop loading in time. Maybe somewhere else is leaking? Perhaps a problem in mpv itself? It should be investigated.

### DIFF
--- a/media_kit/lib/src/libmpv/player.dart
+++ b/media_kit/lib/src/libmpv/player.dart
@@ -1686,7 +1686,7 @@ class Player extends PlatformPlayer {
   final Completer<Pointer<generated.mpv_handle>> _handle =
       Completer<Pointer<generated.mpv_handle>>();
 
-  // [Completer] that signifies that `MPV_EVENT_SHUTDOWN`
+  // [Completer] that signifies that the `MPV_EVENT_SHUTDOWN` event was received. 
   final Completer<void> _mpvShutdownComplete = Completer<void>();
 
   /// A simple flag to prevent changes to [state.playing] due to `loadfile` commands in [open].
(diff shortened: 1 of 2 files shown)


media_kit/lib/src/libmpv/player.dart
```diff
@@ -74,13 +74,29 @@ class Player extends PlatformPlayer {
...
+
+      // Wait for shutdown event to actually occur. 
+      // If timed out, don't do anything. Perhaps implement a solution to handle this in the future.
+      await _mpvShutdownComplete.future.timeout(Duration(seconds: 3), onTimeout: () {});
+
+      // On android, calling `quit 0` does not actually clean the resources.
+      // We must call `mpv_destroy()` in order to prevent extremely high CPU usage.
+      // In the future, find out whether other platforms require this as well.
+      if(Platform.isAndroid){
+        _libmpv?.mpv_destroy(ctx);
```
birros, line 97: Doesn't seem to crash on macOS, if it's helpful for Android it should be helpful for other platforms too.

alexmercerind:
Dealt as part of #189. A more reliable & safer approach; that for all platforms.

PR #199, opened 2023-05-20
Create GL context from the parent window of the FlView widget

### LINKED ISSUE #198: Hardware acceleration not available when running on Ubuntu 22.04
Hardware acceleration is disabled when running on Ubuntu 22.04 with X window system, with GDK compalining that:
```
Unable to create a GL context.
```

After some digging, I found that the window returned by `gdk_get_default_root_window` is different from the window created by flutter, causing GDK fails to create GL context for the wrong window.

So if I pass the `FlView*` into `video_output_new` and call `gdk_window_create_gl_context` with `gtk_widget_get_window(GTK_WIDGET(fl_view))`, everything works fine.

Side note:

As for Wayland on Ubuntu 22.04, media_kit claims that hardware accceleration is enabled but it's actually consuming a **lot** of CPU (showing 700% plus on my PC), which is much worse than software rendering.

noelex:
This PR enables hardware acceleration support for Ubuntu 22.04 with X window system, fixes #198.

Please note that the high CPU consumption issue when running on Wayland is not addressed by this PR.


alexmercerind:
Thanks again!

PR #206, opened 2023-05-24
feat: iOS Picture in Picture support

23doors:
PIP with texture API is only available since iOS 15+ so this is a must.

To enable PIP app needs to have audio background mode enabled in Info.plist:
```
	<key>UIBackgroundModes</key>
	<array>
		<string>audio</string>
	</array>
```

and that's pretty much it. New APIs:

```
controller.isPictureInPictureAvailable() -> only returns true on iOS15+ platform
controller.enablePictureInPicture()
controller.disablePictureInPicture()
controller.enableAutoPictureInPicture() -> when app goes into background, currently playing video will automatically go into PiP
controller.disableAutoPictureInPicture()
controller.enterPictureInPicture() -> requires PiP to be enabled first (or auto PIP)
```

### DIFF
--- a/media_kit/lib/src/models/player_streams.dart
+++ b/media_kit/lib/src/models/player_streams.dart
@@ -83,6 +83,9 @@ class PlayerStreams {
   /// Currently playing video's height.
   final Stream<int> height;
 
+  /// Fired on video position seek.
+  final Stream<Duration> seek;
+
   /// {@macro player_streams}
   const PlayerStreams(
     this.playlist,
@@ -105,5 +108,6 @@ class PlayerStreams {
     this.tracks,
     this.width,
     this.height,
+    this.seek,
   );
 }
--- a/media_kit/lib/src/player/libmpv/player/real.dart
+++ b/media_kit/lib/src/player/libmpv/player/real.dart
@@ -523,6 +523,8 @@ class libmpvPlayer extends PlatformPlayer {
         args.cast(),
       );
       calloc.free(args);
+
+      seekController.add(duration);
     }
 
     if (synchronized) {
--- a/media_kit/lib/src/player/platform_player.dart
+++ b/media_kit/lib/src/player/platform_player.dart
@@ -61,6 +61,7 @@ abstract class PlatformPlayer {
     tracksController.stream,
     widthController.stream,
     heightController.stream,
+    seekController.stream,
   );
 
   @mustCallSuper
@@ -87,6 +88,7 @@ abstract class PlatformPlayer {
         tracksController.close(),
         widthController.close(),
         heightController.close(),
+        seekController.close(),
       ],
     );
     for (final callback in release) {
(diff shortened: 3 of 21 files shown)


birros:
I just tested it, awesome!

However, I have encountered some bugs:
1. sometimes the PIP displays a black screen, especially when I switch to network videos, but I have trouble reproducing
2. using software rendering when the app is unfocused causes colorimetry problems with HDR videos, but this problem also exists with non-PIP software rendering. @23doors are you sure there is no way to keep hardware rendering when the app is unfocused?

---

Finally, there is still some work to do before we can merge:
1. @alexmercerind needs to confirm that he's ok with adding this kind of API to the `VideoController`, maybe review the syntax
2. certainly review the `VideoOutput.swift` code, probably by moving it to a dedicated file…

23doors:
Regarding 1. Does it fix itself or stays black? I can try to fix it if it is reproducable but as you know, I have near zero experience with MacOS/iOS and swift in general.

Regarding 2. Yes sadly this seems to be the case. Any kind of opengl calls when app is in background cause a bunch of warnings about app not having access to GPU. This is also consistent with what I found in docs about iOS background mode.

https://developer.apple.com/documentation/metal/gpu_devices_and_work_submission/preparing_your_metal_app_to_run_in_the_background
>iOS and tvOS restrict a background app’s access to the GPU, to guarantee foreground app performance.

Regarding review.
Feel free to suggest whatever you feel is necessary. This is pretty much first time I touched swift. Separated VideoOutputPIP.swift for now

birros:
So, after 2 weeks of considerations, I'm not OK to merge this PR until a solution is found for `HEVC` videos rendering in PIP:
- by finding a way to enable the GPU when the app is in background
- or by fixing colors in software rendering

If we merge with this "bug", people will think this feature is ready, and will lead to countless issues about it…

@23doors sorry, that doesn't stop you from keeping your branch up to date, until we find a solution, or someone help us…

@alexmercerind I set this PR as `draft` with the `help-wanted` label if you agree.

alexmercerind:
Hi! I apologize for the delay.

I'm still thinking over the code style. Ideally, we'd land this after the first-stable 1.0.0 release, once it's well-tested & (hopefully) implemented for other platforms too.

kkkkkkkkkkkkkkeee:
Excuse me, everyone. Is this still going on

AlvaroVasconcelos:
I really need this functionality, is there anything new?

andynvt:
yes, really need it

> I really need this functionality, is there anything new?

Yasinilhan61:
@alexmercerind any update about this?

kalismeras61:
@alexmercerind could you spare some time for this? This is an important feature and I believe it will differentiate this plugin from other plugins.

Hilbert2048:
@alexmercerind  any update about this?

Hilbert2048:
@birros any update about this?

PR #232, opened 2023-06-10
fix: default volume 1 to 100 in player state

zezo357:
Changing the default volume in player state from 1 to 100,


alexmercerind:
This issue was resolved internally (with appropriate tests; will link them here once public). Merging for the sake of completion.

PR #246, opened 2023-06-19
fix: buffering not true during starting of stream playback

zezo357:
switched to core-idle instead of paused-for-cache, for buffering state

a fix for https://github.com/alexmercerind/media_kit/issues/227

### DIFF
--- a/media_kit/lib/src/player/libmpv/player/real.dart
+++ b/media_kit/lib/src/player/libmpv/player/real.dart
@@ -1183,6 +1183,7 @@ class libmpvPlayer extends PlatformPlayer {
           prop.ref.format == generated.mpv_format.MPV_FORMAT_FLAG) {
         final buffering =
             (prop.ref.data.cast<Int8>().value != 0 && state.playing);
+        print("buffering: " + buffering.toString());
         state = state.copyWith(buffering: buffering);
         if (!bufferingController.isClosed) {
           bufferingController.add(buffering);
@@ -1484,14 +1485,15 @@ class libmpvPlayer extends PlatformPlayer {
         if (value) {
           if (isPlayingStateChangeAllowed) {
             state = state.copyWith(
-              playing: false,
-              completed: true,
-            );
+                playing: false, completed: true, buffering: false);
             if (!playingController.isClosed) {
               playingController.add(false);
             }
             if (!completedController.isClosed) {
               completedController.add(true);
+            }
+            if (!bufferingController.isClosed) {
+              bufferingController.add(false);
             }
           }
         }
--- a/media_kit/lib/src/player/libmpv/player/real.dart
+++ b/media_kit/lib/src/player/libmpv/player/real.dart
@@ -1136,15 +1136,16 @@ class libmpvPlayer extends PlatformPlayer {
 
     if (event.ref.event_id == generated.mpv_event_id.MPV_EVENT_START_FILE) {
       if (isPlayingStateChangeAllowed) {
-        state = state.copyWith(
-          playing: true,
-          completed: false,
-        );
+        state =
+            state.copyWith(playing: true, completed: false, buffering: true);
         if (!playingController.isClosed) {
           playingController.add(true);
         }
         if (!completedController.isClosed) {
           completedController.add(false);
+        }
+        if (!bufferingController.isClosed) {
+          bufferingController.add(true);
         }
       }
     }
@@ -1183,7 +1184,6 @@ class libmpvPlayer extends PlatformPlayer {
           prop.ref.format == generated.mpv_format.MPV_FORMAT_FLAG) {
         final buffering =
             (prop.ref.data.cast<Int8>().value != 0 && state.playing);
-        print("buffering: " + buffering.toString());
         state = state.copyWith(buffering: buffering);
         if (!bufferingController.isClosed) {
           bufferingController.add(buffering);
--- a/media_kit/lib/src/player/libmpv/player/real.dart
+++ b/media_kit/lib/src/player/libmpv/player/real.dart
@@ -1136,17 +1136,20 @@ class libmpvPlayer extends PlatformPlayer {
 
     if (event.ref.event_id == generated.mpv_event_id.MPV_EVENT_START_FILE) {
       if (isPlayingStateChangeAllowed) {
-        state =
-            state.copyWith(playing: true, completed: false, buffering: true);
+        state = state.copyWith(
+          playing: true,
+          completed: false,
+        );
         if (!playingController.isClosed) {
           playingController.add(true);
         }
         if (!completedController.isClosed) {
           completedController.add(false);
         }
-        if (!bufferingController.isClosed) {
-          bufferingController.add(true);
-        }
+      }
+      state = state.copyWith(buffering: true);
+      if (!bufferingController.isClosed) {
+        bufferingController.add(true);
       }
     }
     // NOTE: Now, --keep-open=yes is used. Thus, eof-reached property is used instead of this.
@@ -1173,7 +1176,7 @@ class libmpvPlayer extends PlatformPlayer {
       if (prop.ref.name.cast<Utf8>().toDartString() == 'pause' &&
           prop.ref.format == generated.mpv_format.MPV_FORMAT_FLAG) {
         if (isPlayingStateChangeAllowed) {
-          final playing = prop.ref.data.cast<Int8>().value != 1;
+          final playing = prop.ref.data.cast<Int8>().value == 0;
           state = state.copyWith(playing: playing);
           if (!playingController.isClosed) {
             playingController.add(playing);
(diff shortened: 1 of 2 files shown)


alexmercerind:
If possible, can we have some unit-test(s) for this (no need to modify existing)?

This is one of the remaining things which are still un-tested. PlaylistMode` & `add`/`remove`/`move` methods are still un-tested, which I'm taking care of... alongside web implementation parity checking).

alexmercerind:
As far as I can see at the moment, I think these tests are enough:
- player-buffering-open-file
  - Check buffering emits/does-not-emit if a file:// is opened.
- player-buffering-open-network
  - Check buffering emits if an http:// is opened.
- player-buffering-false-upon-complete.
  - Check buffering is restored to false upon completion of media.
- player-buffering-upon-seek
  - Check buffering state is entered upon seeking out of "buffered" area.
  - [UNSURE] May require local `HttpServer` like headers test(s) to simulate.

alexmercerind:
Existing `player-open-playable-*` tests check `playing`/`completed`/`playlist` `Stream`(s) very strictly.
I think bringing `buffering` here will bring unnecessary complexity. It's enough to check buffering separately.

zezo357:
> As far as I can see at the moment, I think these tests are enough:
> - player-buffering-open-file
>   - Check buffering emits/does-not-emit if a file:// is opened.
> - player-buffering-open-network
>   - Check buffering emits if an http:// is opened.
> - player-buffering-false-upon-complete.
>   - Check buffering is restored to false upon completion of media.
> - player-buffering-upon-seek
>   - Check buffering state is entered upon seeking out of "buffered" area.
>   - [UNSURE] May require local `HttpServer` like headers test(s) to simulate.
> 

Thank you, will check them tomorrow and let you know when I finish it

zezo357:
hello, i am having problems writing the tests 
logic:
- when i used the core idle i was checking if idle and state is playing (if not then its buffering)
problem:
- thats is not valid for completed (since the idle call comes before eof the player states that its buffering when it should be not)
- writing tests: the buffering is never true (i guess since there is no ui not sure)


fixes:
- when the player is completed overwrite buffering with false, dont know if thats valid or not and if it will cause other problems
- i guess a full integration tests will simulate real application behaviour.

now i have stated all the problems, i will try to fix them and hope thats fix is valid (but i still hate the fake buffering call when completed in my fix)

zezo357:
i did apply the fix, but still the buffering is never true in the tests, meanwhile it works on the app,

and i cant fix that, so the tests are never true and that shouldn't happen, will push the fix and the state of the current tests 

but i need help in that problem

alexmercerind:
We'd still require tests before landing any change.

zezo357:
> We'd still require tests before landing any change.
> 
> If it works as intended, tests shall pass.

Well in this case the tests are not valid if they don't show rendering since they don't buffer,

But in a real app they do,


I tried to make integration tests but couldn't since I never wrote them,

alexmercerind:
They are valid.

zezo357:
> They are valid.

tests a working correctly now, also i added a network url (for a 100+ mb video) for the buffering to be valid on seek

i didnt add it to sources since it will download it each time it ran it, which is too much for my internet, and i dont need it downloaded

somehow it fails on github :) i will try to fix it :)

zezo357:
the tests are all passing now, finally :)

media_kit/lib/src/player/libmpv/player/real.dart
```diff
@@ -1136,16 +1136,17 @@ class libmpvPlayer extends PlatformPlayer {
...
+        state =
+            state.copyWith(playing: true, completed: false, buffering: true);
         if (!playingController.isClosed) {
           playingController.add(true);
         }
         if (!completedController.isClosed) {
           completedController.add(false);
         }
+        if (!bufferingController.isClosed) {
+          bufferingController.add(true);
```
23doors, line 1148: @zezo357 Did you test if maybe this change alone is enough?

I mean. From what I checked, when using `paused-for-cache` you still get events for buffering=false once buffering is done. So emitting initial buffering=true after file is starting to load seems to be enough and no need to switch to `core-idle`. Isn't core-idle switching to true also when video is paused (so it would be misleading to consider that buffering)?

zezo357: from my testing paused-for-cache is never true even when buffering 

and yes core-idle is true when video is paused, so i did make the buffering a combination of (idle and playing) so it will be only true if idle for buffering

zezo357: also since you mentioned that, i will check if paused-for-cache can be used, and valid using the current tests

zezo357: As you can see, the tests fails 

And seek doesn't buffer even though it should

23doors: Weird behavior on mpv side, but nice workaround for it. Thanks for checking!

zezo357: Thank you for reviewing the PR.

alexmercerind: The `pause-for-cache` gets `true` (as intended) e.g. internet gets slow/disconnected in the middle of playback.

The only issue with it was that it never worked during "starting" i.e. initial buffering of the stream. That's one thing we're looking for.

zezo357:
> Still doesn't work right IMO. It shouldn't be true again upon completion.
> 
> As far as I can understand, player-buffering-false-upon-complete does "nothing". I have to test things out.

yeah my workaround makes it true and then false upon completion, cant figure out a better workaround

media_kit/lib/src/player/libmpv/player/real.dart
zezo357: > The `pause-for-cache` gets `true` (as intended) e.g. internet gets slow/disconnected in the middle of playback... and position crosses the buffered zone.
> 
> The only issue with it was that it never worked during "starting" i.e. initial buffering of the stream. That's one thing we're looking for.

well as you can see in this https://github.com/zezo357/media_kit/actions/runs/5356947170 it never goes true, same for the issue https://github.com/alexmercerind/media_kit/issues/227

alexmercerind: It goes true, we're using it... only when it pauses for buffering, not at the start of stream.

alexmercerind:
You've left the future events untested... Who knows what's happening afterward, it is just commented out.

We want to test until `streamDone`.

media_kit/lib/src/player/libmpv/player/real.dart
zezo357: > It goes true, we're using it... only when it pauses for buffering, not at the start of stream.

well that didnt happen in the test mentioned and in the issue

alexmercerind: I've built this; I know better :)

zezo357:
> You've left the future events untested... Who knows what's happening afterward, it is just commented out.
> 
> We want to test until `streamDone`.

what do you mean by streamDone?

alexmercerind:
`emitsDone`

media_kit/lib/src/player/libmpv/player/real.dart
zezo357: > 

want me to switch back to `pause-for-cache` ? and change logic back to it? even though that from my tests/ usage of the package the buffering was never true in my case (it will pause to buffer but the buffering is not true) keep in mind this is for local files too 

like after seek it will pause for a small bit without buffering `true` in this pr the buffering will be true

alexmercerind (approved):
LGTM!

I noticed that `player.open(..., play: false)` was not triggering buffering as intended. The default, `player.open(...)` _i.e._ `player.open(..., play: true)` triggered buffering correctly.

I fixed the issue & added `player-buffering-*-play-false` test(s). Other tests have also been improved to run till end-of-file & `player.dispose()`; primarily because we need to track the stream's behavior.

I'm merging this into #250 tree; because of required updates & web implementation also needs to be tested against these tests. The GitHub UI will automatically "merge" once #250 lands.

<hr>

The core idea behind adding unit-tests is:
- Stability (obviously)
- Avoiding regressions as dependents may use any behavior in their code.
- Preventing breaking changes when updating libmpv & other upstream packages etc.
- Maintaining parity & similarity between native implementation & web implementation.

I've been very busy with #250; a ton of features & fixes + many more unit-tests (take over 25 minutes to run). It may not seem like it but providing support & talking to people here takes a lot of time & most importantly breaks the "flow".

<hr>

alexmercerind:
Hopefully the tests also passed.

alexmercerind:
Same buffering is implemented for the web & one more test:

https://github.com/alexmercerind/media_kit/blob/7c320f656014a0adc844b0274f83adb78a9e3a98/media_kit/test/src/player/player_test.dart#L2366-L2435

media_kit/lib/src/player/libmpv/player/real.dart
```diff
@@ -1172,16 +1176,24 @@ class libmpvPlayer extends PlatformPlayer {
...
+          final playing = prop.ref.data.cast<Int8>().value == 0;
           state = state.copyWith(playing: playing);
           if (!playingController.isClosed) {
             playingController.add(playing);
           }
         }
       }
+      if (prop.ref.name.cast<Utf8>().toDartString() == 'core-idle' &&
+          prop.ref.format == generated.mpv_format.MPV_FORMAT_FLAG) {
+        final buffering = prop.ref.data.cast<Int8>().value == 1;
```
zezo357, line 1188: should this be `and playing`?

since if this condition doesn't exist it will buffering is true when paused?

alexmercerind: I noticed the issue & resolved it in #250 tree. The `&& playing` had some race issue which made it fail on runner.

https://github.com/alexmercerind/media_kit/blob/e2588461bd8c0a378379f8af911587cb3463e552/media_kit/lib/src/player/libmpv/player/real.dart#L1242-L1255

PR #282, opened 2023-07-12
feat: external subtitle track & audio track support

alexmercerind:
- feat: expose `SubtitleTrack.external` & `AudioTrack.external`
- refactor: `libmpvPlayer` -> `NativePlayer` & `webPlayer` -> `WebPlayer`


alexmercerind:
Requesting a review.

On web, one can load WebVTT as `String` directly as-well:

https://github.com/alexmercerind/media_kit/blob/3adca06df1f6dc565ab6e3ba0306441ea4681eb5/media_kit/test/src/player/player_test.dart#L3124-L3375

I felt it is important since loading WebVTT subtitles from other domains is restricted (CORS).

A better approach could've been adding a new factory constructor to `SubtitleTrack` for loading external subtitle as `String` to the native backend too. But, in the end we would have to save it to a temporary file & then feed it into libmpv + how many people are going to use it. This would also require a (massive at the moment) of fetching temporary directory (& permissions?) in our Dart-only root media_kit package.

<hr>

The tests should pass (probably).

birros:
> On web, one can load WebVTT as String directly as-well:
> I felt it is important since loading WebVTT subtitles from other domains is restricted (CORS).

I think it's a bad idea for two reasons:
1. It's confusing because only implemented on the web…
2. …since internally it only takes a few lines to create a temporary URL from a raw string, an example in the documentation to web developers is enough

---

In any case, if you plan to keep support for loading subtitles as raw string, you need to add a mechanism for revoking created urls, otherwise this will cause memory leaks:

```dart
final blob = html.Blob([webVTTString]);
final url = html.Url.createObjectUrlFromBlob(blob);
html.Url.revokeObjectUrl(url);
```

birros (approved):

PR #287, opened 2023-07-17
build: bump darwin libs versions to v0.4.2

birros:
- bump min iOS version to `9.0`
- add `--no-video` when `vo` is `null` and no `VideoController` is used


media_kit/lib/src/player/native/player/real.dart
```diff
@@ -2001,6 +2001,13 @@ class NativePlayer extends PlatformPlayer {
       // The options which must be set before [MPV.mpv_initialize].
       final options = <String, String>{};
 
+      // Disable video ouput by default, overwritten by [VideoController]
```
alexmercerind, line 2004: nit: typo

alexmercerind:
I hope this looks good to you.

alexmercerind:
I just disabled macOS unit-tests; I don't know why `player-subtitle` is failing. Sure, feel free if you have clues.

birros:
What you've done works, there are no crash with audio libraries because `vid` is set to `no`.

However, I think keeping it by default as I did when no `VideoController` is present is relevant in one case.
If someone wants to use a player only for audio session when the video libraries are used.
Without it, video decoding is still active and can have an impact on battery life.

Maybe I'm wrong and it has no impact on the battery, maybe mpv just decodes the headers…

PR #291, opened 2023-07-19
perf: vid=no until `VideoController` attach; build(windows): bump native libraries

alexmercerind:
Re-work things based on https://github.com/alexmercerind/media_kit/pull/287#issuecomment-1641110217.

Additionally handle a condition if media_kit_libs_***_audio are used with `VideoController` (audio still plays):


birros (approved):
LGTM!

Perfect! Good idea to keep the warning in `VideoController`.

PR #292, opened 2023-07-20
build: bump media_kit_libs_* version(s)

alexmercerind:


birros:
@alexmercerind From Discord:
> If you notice I actually bumped platform back to 13.0 in iOS. 
> This should probably reduce the resulting iOS IPA size again. 
> As far as I undersand... probably old iOS versions (< 13.0) don't have (or have old) Swift runtime libraries (as part of OS). 

Nice catch! You're absolutely right, XCode adds a lot of polyfill to be retro compatible.
I just fixed the `Podfile` you edited, and changed the min version in the `project.pbxproj`.

---

Unzipped `build/ios/iphoneos/Runner.app`:
```
iOS min version: 11.0  -> 13.0
Size:            218MB -> 57MB
```

README.md
birros, line 1: I think that reverts your last README modification, doesn't it?

media_kit/README.md
birros, line 1: Same here.

birros (approved):
LGTM!

PR #293, opened 2023-07-21
feat: improve wakelock behavior

birros:
- Sync wakelock state with playing state
- Manage concurrent wakelock calls

### DIFF
--- a/media_kit_video/lib/src/utils/wakelock.dart
+++ b/media_kit_video/lib/src/utils/wakelock.dart
@@ -1,0 +1,42 @@
+import 'package:wakelock_plus/wakelock_plus.dart';
+
+/// Manage the wakelock behavior for multiple instances.
+class Wakelock {
+  static final Set<Wakelock> _enabledInstances = {};
+  static bool _wakelockEnabled = false;
+
+  /// Marks the wakelock as enabled for this instance.
+  void enable() {
+    _enabledInstances.add(this);
+    _updateWakeLock();
+  }
+
+  /// Marks the wakelock as disabled for this instance.
+  void disable() {
+    _enabledInstances.remove(this);
+    _updateWakeLock();
+  }
+
+  /// Toggles the wakelock status for this instance.
+  void toggle() {
+    if (_enabledInstances.contains(this)) {
+      _enabledInstances.remove(this);
+    } else {
+      _enabledInstances.add(this);
+    }
+    _updateWakeLock();
+  }
+
+  /// Updates the upstream wakelock status based on the enabled instances count.
+  static void _updateWakeLock() {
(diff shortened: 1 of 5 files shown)


media_kit_video/lib/src/video/video_texture.dart
```diff
@@ -133,19 +138,39 @@ class VideoState extends State<Video> {
...
+      duration: duration,
+    );
+  }
+
   @override
   void initState() {
     super.initState();
     if (widget.wakelock) {
-      WakelockPlus.enable().catchError((_) {});
+      if (widget.controller.player.state.playing) {
```
alexmercerind, line 155: I think we must acquire wake-lock if `Player` is already in `playing` state before `Video` mounts.

media_kit_video/lib/src/utils/wakelock.dart
```diff
@@ -0,0 +1,53 @@
...
+      WakelockPlus.disable().catchError((_) {});
+    }
+    debugPrint('media_kit: wakelock: _count = $_count');
+  }
+
+  /// Whether the wakelock is enabled for this instance.
+  bool _enabled = false;
+
+  /// The number of [Wakelock] instances in enabled state.
+  static int _count = 0;
```
alexmercerind, line 52: I find this count-based implementation bit simpler to understand.

A single static field & we don't need to store instances inside a set.

alexmercerind (approved):
LGTM! Thanks for having a look.

I hope you are fine with the changes; or any other thing if I missed.

media_kit_video/lib/src/utils/wakelock.dart
birros: Hmm, this class is useless ^^
You can delete it, because `WakelockPlus` does the same thing, but statically...

My first implementation was based on counters, but it has side-effects, as counters often do.
If `enable` or `disable` is called more than once then the counter becomes inconsistent, for example if the video is paused before being disposed.

alexmercerind: I think you are probably reading it wrong, each instance can only cause an increment of 1 to the `_count`.

https://github.com/alexmercerind/media_kit/blob/c7b46bfd38c4b4311e049d82286e4b74903bee17/media_kit_video/lib/src/utils/wakelock.dart#L22
https://github.com/alexmercerind/media_kit/blob/c7b46bfd38c4b4311e049d82286e4b74903bee17/media_kit_video/lib/src/utils/wakelock.dart#L31

birros: Yep, you are right. Sorry. I didn’t see the static counter…

PR #294, opened 2023-07-21
feat/refactor: SubtitleTrack.data support in Player

alexmercerind:
```dart
await player.open(
  Media(
    'https://www.iandevlin.com/html5test/webvtt/v/upc-tobymanley.theora.ogg',
  ),
);
await player.setSubtitleTrack(
  SubtitleTrack.data(
    '''WEBVTT FILE

1
00:00:03.500 --> 00:00:05.000 D:vertical A:start
Everyone wants the most from life

2
00:00:06.000 --> 00:00:09.000 A:start
Like internet experiences that are rich <b>and</b> entertaining

3
00:00:11.000 --> 00:00:14.000 A:end
Phone conversations where people truly <c.highlight>connect</c>

4
00:00:14.500 --> 00:00:18.000
Your favourite TV programmes ready to watch at the touch of a button

5
00:00:19.000 --> 00:00:24.000
Which is why we are bringing TV, internet and phone together in <c.highlight>one</c> super package

6
00:00:24.500 --> 00:00:26.000
<c.highlight>One</c> simple way to get everything

7
00:00:26.500 --> 00:00:27.500 L:12%
UPC

8
00:00:28.000 --> 00:00:30.000 L:75%
Simply for <u>everyone</u>
''',
    title: 'English',
    language: 'en',
  ),
);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -781,12 +781,57 @@ https://user-images.githubusercontent.com/28951144/253067794-73b5ca5d-e90d-4892-
 
 ### Load external subtitle track
 
-The `SubtitleTrack.external` constructor can be used to load external subtitle track with URI e.g. SRT, WebVTT etc. The code is easier to understand:
+The `SubtitleTrack.uri` constructor can be used to load external subtitle track **with URI** e.g. SRT, WebVTT etc. The code is easier to understand:
 
 ```dart
 await player.setSubtitleTrack(
-  SubtitleTrack.external(
+  SubtitleTrack.uri(
     'https://www.iandevlin.com/html5test/webvtt/upc-video-subtitles-en.vtt',
+    title: 'English',
+    language: 'en',
+  ),
+);
+```
+
+The `SubtitleTrack.data` constructor can be used to load external subtitle track **with data** e.g. SRT, WebVTT etc. The code is easier to understand:
+
+```dart
+player.setSubtitleTrack(
+  SubtitleTrack.data(
+    '''WEBVTT FILE
+
+1
+00:00:03.500 --> 00:00:05.000 D:vertical A:start
+Everyone wants the most from life
+
+2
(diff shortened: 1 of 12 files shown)


birros (approved):
LGTM!

birros:
> A cleaner approach can be to add a new factory constructor to SubtitleTrack to split away URI parsing logic. WDYT?

I think it's better.
As I told you in another PR, the fact that the Player itself creates temporary files or blobs without deleting them can lead to memory leaks.
So having a different constructor allows us to differentiate it, even in the documentation, by adding a warning about memory leaks.

alexmercerind:
Re-requesting review.

- [Now `SubtitleTrack.uri` & `SubtitleTrack.data` exist independently & work on all platforms/backends.](https://github.com/alexmercerind/media_kit/tree/feat/external-subtitle-string#load-external-subtitle-track)

> As I told you in another PR, the fact that the Player itself creates temporary files or blobs without deleting them can lead to memory leaks.

I forgot to update you in the last PR, we're not leaking memory, unlinking the temporary file (native) & revoking object URL (web) upon `dispose). I don't think a subtitle string is going to be _that_ large anyway (when compared to a binary file).

https://github.com/alexmercerind/media_kit/blob/126c3d4cffb8fc187bb9ecfae74df22e49f4839f/media_kit/lib/src/player/native/player/real.dart#L1226-L1227
https://github.com/alexmercerind/media_kit/blob/126c3d4cffb8fc187bb9ecfae74df22e49f4839f/media_kit/lib/src/player/web/player/real.dart#L1293-L1296

Regardless, users are free to create the temporary-file/object-URL from subtitle data & then load that URI manually. This will provide (probably) better control, though I believe what we're offering is great.

birros (approved):
LGTM!

PR #300, opened 2023-07-23
Do not add onTapUp callback if toggleFullscreenOnDoublePress is disabled

kirill-21:
This allows you to wrap your Video widget with additional GestureDetector to add onTap callbacks. Right now if you wrap this widget with GestureDetector then you won't be able to execute any onTap callbacks. Of course this could be done with Listener widget, but then it will cover whole player, even play/resume buttons, etc., what's not good


alexmercerind (approved):
LGTM!

alexmercerind:
I just noticed... I forgot to implement `modifyVolumeOnScroll` the other day.

PR #319, opened 2023-08-03
fix: memory leaks

alexmercerind:


alexmercerind:
cc @birros:

I think we should update the macOS & iOS side too.
Make sure check for `MPV_FORMAT_NODE_MAP` otherwise you may encounter crash.

birros:
> I think we should update the macOS & iOS side too.

It's already done: https://github.com/media-kit/media-kit/blob/e5ab431183c81272e0a5545fb791273d751dbd8e/media_kit_video/common/darwin/Classes/plugin/MPVHelpers.swift#L13-L15

`defer` means that at the end of the method whatever is inside is called.

---

> Make sure check for MPV_FORMAT_NODE_MAP otherwise you may encounter crash.

Here too: https://github.com/media-kit/media-kit/blob/e5ab431183c81272e0a5545fb791273d751dbd8e/media_kit_video/common/darwin/Classes/plugin/MPVHelpers.swift#L19-L21

birros (approved):
LGTM!

PR #328, opened 2023-08-05
feat: Improve model exports

AlexV525:
- Exports `PlayerState` and `PlayerStream` since they are public APIs and their `Type` cannot be referenced outside. (e.g. `final PlayerState state = player.state;`)
- Sort exports alphabetically.

### DIFF
--- a/media_kit/lib/media_kit.dart
+++ b/media_kit/lib/media_kit.dart
@@ -7,17 +7,20 @@
 // ignore_for_file: camel_case_types
 
 export 'package:media_kit/src/media_kit.dart';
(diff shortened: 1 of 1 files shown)


alexmercerind:
Thanks for the PR!

The change doesn't offend anyone. I didn't expect people to use the composition `class`es.

AlexV525:
Imagine you have a brdging project that needs to bridge multiple basic values to another platform, what we've been trying to do is writing such code:
```dart
Object? _getStateById(
  String id,
  Object(PlayerState? state) Function onState,
) => onState(_players[id]?.state);
```

PR #330, opened 2023-08-05
perf: move `InitializerNativeEventLoop` calls to another `Isolate`

alexmercerind:


alexmercerind:
I could feel a significant jank after existing current "stress test" screen (because all `Player`s would get `dispose`'d at once).

Issue #339, opened 2023-08-06
[Android] First frame is sometimes a solid color


deckerst:
Some video files loaded on Android show a frame of a solid color instead of the first frame of the video. It is easier to see by opening the video with `play: false`.

You have indicated this delay:
https://github.com/media-kit/media-kit/blob/5ea2c84b5c7b074a45b3887dc64a997f9a5e3172/media_kit_video/lib/src/video_controller/android_video_controller/real.dart#L146-L148

I played with it and increased it to 2000 ms but it does not solve the issue.

I've tried the `paint_first_frame.dart` page of the test app, and everything was fine with the default sources, showing the correct frame for each video. However, when I switched the sources to some local files on storage (not dynamically downloaded), I noticed the solid color frame issue. Easy to reproduce with both `content://...` URIs and regular paths like `'/storage/emulated/0/Pictures/...`.

Interestingly, if we load the page again and again in the test app using local sources, sometimes some of the videos will load correctly, so it definitely points to a race condition. It just does not seem to be related to the delay mentioned above.

<img width="200" src="https://github.com/media-kit/media-kit/assets/13898333/8d8acbd8-1934-4356-b1b3-1deaeeeaf745" />
<img width="200" src="https://github.com/media-kit/media-kit/assets/13898333/0ade8a3e-01c0-4210-b2f2-09debb50310d" />
<img width="200" src="https://github.com/media-kit/media-kit/assets/13898333/6a1ab484-198d-4ccd-8460-193852b9d9a7" />
<img width="200" src="https://github.com/media-kit/media-kit/assets/13898333/383bdd16-d2ce-4d69-b6f0-32dbb994147f" />

alexmercerind:
Hi! This has been fixed with the latest update, please refer to the "Installation" section in the README.

PR #342, opened 2023-08-07
fix(android): SurfaceTexture.setDefaultBufferSize & render race; fix(android): use hwdec=auto; feat(android): VideoControllerConfiguration.scale;

alexmercerind:
- feat(android): `VideoControllerConfiguration.scale`
- fix(android): use `hwdec=auto`
- fix(android): `SurfaceTexture.setDefaultBufferSize` & render race

### DIFF
--- a/media_kit/lib/src/player/native/player/real.dart
+++ b/media_kit/lib/src/player/native/player/real.dart
@@ -2082,6 +2082,8 @@ class NativePlayer extends PlatformPlayer {
         // OpenSL ES audio output seems to be broken on some of these.
         else if (AndroidHelper.isEmulator && AndroidHelper.APILevel <= 25)
           'ao': 'null',
+        'subs-fallback': 'yes',
+        'subs-with-matching-audio': 'yes',
       };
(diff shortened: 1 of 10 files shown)


alexmercerind:
cc @birros:

At the moment, we provide `width` & `height` to limit the video output size for yielding higher performance on certain devices. While the feature is good to have, it has a few issues if used from a performance standpoint:
- Not every video will have the same resolution or require downscaling (or few may even get redundantly "upscaled").
- Having a fixed width/height may result in the wrong aspect-ratio for different videos.

It also didn't work on Android because `android.view.Surface` aspect-ratio must be equal to source aspect-ratio.

<hr>

I have added this new API/feature: `VideoControllerConfiguration.scale`. This allows changing the video output size independent of the video's actual resolution (without wrongly affecting the aspect-ratio).

I gave a thought to the naming, I believe `scale` fits the purpose. Initially I thought about `pixel-ratio`, but that is entirely different thing according to the Internet.

alexmercerind:
Closes #339.

PR #350, opened 2023-08-09
feat: `NativePlayer.observeProperty` & `NativePlayer.unobserveProperty`

alexmercerind:
A new forbidden API that might be useful in some advanced use-cases. I have added relevant unit-tests.

**Example Usage:**

```dart
import 'package:media_kit/media_kit.dart';

Future<void> main() async {
  MediaKit.ensureInitialized();
  final player = Player();
  await (player.platform as NativePlayer).observeProperty(
    'demuxer-cache-state',
    (data) async {
      print(data);
    },
  );
  await player.open(Media('https://user-images.githubusercontent.com/28951144/229373695-22f88f13-d18f-4288-9bf1-c3e078d83722.mp4'));
}
```

**Example Output:**

```
{"cache-duration":0.000000,"eof":false,"underrun":true,"idle":false,"total-bytes":0,"fw-bytes":0,"raw-input-rate":2362,"debug-low-level-seeks":0,"debug-byte-level-seeks":0,"bof-cached":false,"eof-cached":false,"seekable-ranges":[]}
{"cache-end":11.238458,"reader-pts":0.255420,"cache-duration":10.983039,"eof":false,"underrun":false,"idle":false,"total-bytes":448496,"fw-bytes":434496,"raw-input-rate":2362,"debug-low-level-seeks":0,"debug-byte-level-seeks":0,"debug-ts-last":11.238458,"bof-cached":true,"eof-cached":false,"seekable-ranges":[{"start":-0.023220,"end":11.215238}]}
{"cache-end":26.981587,"reader-pts":0.255420,"cache-duration":26.726168,"eof":false,"underrun":false,"idle":false,"total-bytes":1068112,"fw-bytes":1054112,"raw-input-rate":2362,"debug-low-level-seeks":0,"debug-byte-level-seeks":0,"debug-ts-last":26.981587,"bof-cached":true,"eof-cached":false,"seekable-ranges":[{"start":-0.023220,"end":26.958367}]}
{"cache-end":46.045170,"reader-pts":0.255420,"cache-duration":45.789751,"eof":false,"underrun":false,"idle":false,"total-bytes":1820448,"fw-bytes":1806448,"raw-input-rate":2362,"debug-low-level-seeks":0,"debug-byte-level-seeks":0,"debug-ts-last":46.045170,"bof-cached":true,"eof-cached":false,"seekable-ranges":[{"start":-0.023220,"end":46.021950}]}
{"cache-end":60.046803,"reader-pts":0.325079,"cache-duration":59.721723,"eof":true,"underrun":false,"idle":true,"total-bytes":2372112,"fw-bytes":2354976,"raw-input-rate":2362,"debug-low-level-seeks":0,"debug-byte-level-seeks":0,"debug-ts-last":60.046803,"bof-cached":true,"eof-cached":true,"seekable-ranges":[{"start":-0.023220,"end":60.046803}]}
{"cache-end":60.046803,"reader-pts":1.369977,"cache-duration":58.676825,"eof":true,"underrun":false,"idle":true,"total-bytes":2372112,"fw-bytes":2312880,"raw-input-rate":3383889,"debug-low-level-seeks":0,"debug-byte-level-seeks":0,"debug-ts-last":60.046803,"bof-cached":true,"eof-cached":true,"seekable-ranges":[{"start":-0.023220,"end":60.046803}]}
{"cache-end":60.046803,"reader-pts":1.602177,"cache-duration":58.444626,"eof":true,"underrun":false,"idle":true,"total-bytes":2372112,"fw-bytes":2303712,"raw-input-rate":3383889,"debug-low-level-seeks":0,"debug-byte-level-seeks":0,"debug-ts-last":60.046803,"bof-cached":true,"eof-cached":true,"seekable-ranges":[{"start":-0.023220,"end":60.046803}]}
```

In this case, data received in the listener is JSON. It is expected that API consumers will be self-aware of the properties they observe & the type property will have (thus, accordingly handle it as well). This eliminates the need of exposing every single libmpv property/option inside `PlayerState`/`PlayerStream` for internal/public usage.

### DIFF
--- a/media_kit/lib/src/player/native/player/real.dart
+++ b/media_kit/lib/src/player/native/player/real.dart
@@ -1993,6 +1993,11 @@ class NativePlayer extends PlatformPlayer {
               errorController.add(text);
             }
           }
+          if (prefix == 'cplayer') {
+            if (!errorController.isClosed) {
+              errorController.add(text);
+            }
+          }
         }
         // --------------------------------------------------
       }
--- a/media_kit/lib/src/player/native/player/real.dart
+++ b/media_kit/lib/src/player/native/player/real.dart
@@ -1343,6 +1343,71 @@ class NativePlayer extends PlatformPlayer {
     calloc.free(data);
   }
 
+  /// Observes property for the internal libmpv instance of this [Player].
+  /// Please use this method only if you know what you are doing, existing methods in [Player] implementation are suited for the most use cases.
+  ///
+  /// See:
+  /// * https://mpv.io/manual/master/#options
+  /// * https://mpv.io/manual/master/#properties
+  ///
+  Future<int> observeProperty(
+    String property,
+    Future<void> Function(String) listener,
+  ) async {
+    if (disposed) {
+      throw AssertionError('[Player] has been disposed');
+    }
+    await waitForPlayerInitialization;
+    await waitForVideoControllerInitializationIfAttached;
+
+    if (observed.containsKey(property)) {
+      throw ArgumentError.value(
+        property,
+        'property',
+        'Already observed',
+      );
+    }
+    final reply = Random().nextInt(1 << 16) + 1;
+    observed[property] = ObservedProperty(reply, listener);
+    final name = property.toNativeUtf8();
+    mpv.mpv_observe_property(
+      ctx,
+      reply,
+      name.cast(),
+      generated.mpv_format.MPV_FORMAT_NONE,
+    );
+    calloc.free(name);
+    return reply;
+  }
+
+  /// Unobserves property for the internal libmpv instance of this [Player].
+  /// Please use this method only if you know what you are doing, existing methods in [Player] implementation are suited for the most use cases.
+  ///
+  /// See:
+  /// * https://mpv.io/manual/master/#options
+  /// * https://mpv.io/manual/master/#properties
+  ///
+  Future<void> unobserveProperty(String property) async {
+    if (disposed) {
+      throw AssertionError('[Player] has been disposed');
+    }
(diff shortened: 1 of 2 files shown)


media_kit/lib/src/player/native/player/real.dart
```diff
@@ -1343,6 +1343,71 @@ class NativePlayer extends PlatformPlayer {
...
+    await waitForVideoControllerInitializationIfAttached;
+
+    if (observed.containsKey(property)) {
+      throw ArgumentError.value(
+        property,
+        'property',
+        'Already observed',
+      );
+    }
+    final reply = Random().nextInt(1 << 16) + 1;
```
birros, line 1370: I'm going to be the boring one (again) ^^

I'm not a fan of random here, the risk of collision exists, even if it's very small, especially as it's a "hidden" API.

Instead, I would have used the `hashCode` of the property name.
Knowing that the number of properties is relatively low, the probability of collision is even lower.
The other advantage is that if a collision happens, it won't happen randomly but all the time, which will be easier to debug.

In short, it's a small detail, but one that can be a real pain to debug the day it happens (true story).

birros (approved):
LGTM!

media_kit/lib/src/player/native/player/real.dart
alexmercerind: That's awesome feedback.

I don't know why I spent so much time thinking about saving additional `int`. This approach eliminates the need for `ObservedProperty` class entirely. Everything redundant.

The `hashCode` indeed works great for this purpose.

PR #388, opened 2023-08-21
removed check on video player in find packages 

zezo357:
since i am doing what you requested in https://github.com/media-kit/media-kit/pull/382

the check is making my package flutter_meedu_videoplayer crash since its checking for videoplayer

even though its doesnt have media_kit in the name.

for the other packages with media_kit in its name i will change them to another names and disable the old ones.

PR #382, opened 2023-08-19
warning instead of exit

zezo357:
fix for https://github.com/media-kit/media-kit/issues/378

to log a warning instead of exiting the app

PR #407, opened 2023-08-25
chore: remove find packages logic

alexmercerind:
Ref: #382 #388


alexmercerind:
The test "player-buffering-upon-seek" appears to be flaky on GNU/Linux.

Merging.

Issue #393, opened 2023-08-22
[Windows][GNU/Linux] improve CMake script to handle failures better


pufaqi:
- Flutter 3.10.5
- windows
- error:
```
 D:\code\flutter\personal\tss\windows\flutter\ephemeral\.plugin_symlinks\media_kit_video\windows\video_output.h(14,10): fatal  error C1083: �޷��򿪰����ļ�: ��client.h��: No such file or dire ctory     
[D:\code\flutter\personal\tss\build\windows\plugins\media_kit_video\media_kit_video_plugin.vcxproj]
```

pufaqi:
[   +2 ms] executing: [D:\install\flutter/] git -c log.showSignature=false log -n 1 --pretty=format:%H
[  +43 ms] Exit code 0 from: git -c log.showSignature=false log -n 1 --pretty=format:%H
[        ] 796c8ef79279f9c774545b3771238c3098dbefab
[        ] executing: [D:\install\flutter/] git tag --points-at 796c8ef79279f9c774545b3771238c3098dbefab
[  +61 ms] Exit code 0 from: git tag --points-at 796c8ef79279f9c774545b3771238c3098dbefab
[        ] 3.10.5
[   +1 ms] executing: [D:\install\flutter/] git rev-parse --abbrev-ref HEAD
[  +34 ms] Exit code 0 from: git rev-parse --abbrev-ref HEAD
[        ] stable
[ +149 ms] Unable to locate an Android SDK.
[   +2 ms] executing: [D:\install\flutter/] git rev-parse --abbrev-ref --symbolic @{upstream}
[  +35 ms] Exit code 0 from: git rev-parse --abbrev-ref --symbolic @{upstream}
[        ] origin/stable
[        ] executing: [D:\install\flutter/] git ls-remote --get-url origin
[  +39 ms] Exit code 0 from: git ls-remote --get-url origin
[        ] https://github.com/flutter/flutter.git
[  +76 ms] Artifact Instance of 'AndroidGenSnapshotArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'AndroidInternalBuildArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'IOSEngineArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'FlutterWebSdk' is not required, skipping update.
[   +1 ms] Artifact Instance of 'WindowsEngineArtifacts' is not required, skipping update.
[...]

PR #414, opened 2023-08-26
Improving cmake for windows and linux 

zezo357:
https://github.com/media-kit/media-kit/issues/393 hopefully improving the status of this 

to avoid flutter clean when failed 

the logic now will 
check if archive is found -> check its md5 -> if not correct (will delete it) -> then redownload 
if archive is not found -> download it then check its md5


alexmercerind:
Interesting. Surely looks cleaner & definitely improved.

One more case that I believe would be good to handle is that:

[At all places] we're only checking if "directory exists or not", a better check would be to check if "directory is empty or not"
https://github.com/media-kit/media-kit/blob/30121e0717157a30e97d6419591c1bd787b07bda/libs/windows/media_kit_libs_windows_video/windows/CMakeLists.txt#L89

This is mainly because `make_directory` is invoked instantly & the directory is created:

https://github.com/media-kit/media-kit/blob/30121e0717157a30e97d6419591c1bd787b07bda/libs/windows/media_kit_libs_windows_video/windows/CMakeLists.txt#L91

But... the headers & shared libraries etc. are extracted some time afterward (PRE_BUILD), thanks to `add_custom_command`:

If any interruption is made between `make_directory` & `add_custom_command`'s execution, the build will endlessly fail.

https://github.com/media-kit/media-kit/blob/30121e0717157a30e97d6419591c1bd787b07bda/libs/windows/media_kit_libs_windows_video/windows/CMakeLists.txt#L91

<hr>

If one creates an empty `libmpv` or `ANGLE` directory, the build will always fail for no reason (even if the archive exists).

zezo357:
That's a valid point I will check it, hopefully I will come up with a clean solution

If not I will let you know the limition I encountered

Thank you for the review ❤️

zezo357:
done: also tested with angle directory (i deleted all files in it) and it got added again.

alexmercerind (approved):
LGTM!

This is big improvement & will save project from some false issues.

Feel free to merge.

<hr>

One last feedback from my end is to log something as follows whenever integrity check fails:

```
Please try to run "flutter clean" & build project again.
```

Just quitting after saying "integrity check failed." probably seems bit scary to common Flutter developer.

zezo357:
> LGTM!
> 
> This is big improvement & will save project from some false issues.
> 
> Feel free to merge.
> 
> One last feedback from my end is to log something as follows whenever integrity check fails:
> 
> ```
> Please try to run "flutter clean" & build project again.
> ```
> 
> Just quitting after saying "integrity check failed." probably seems bit scary to common Flutter developer.

i think only `build failed (with the error) and please try to build project again.` is enough 
since if they have internet problem and try to download two files one may fail and the other has no problem
so a rebuild will download the only failed one

alexmercerind:
Good point. I was still thinking based on the old CMake script... now that it deletes existing file & redownloads, it seems all right.

alexmercerind (approved):
Re- LGTM!

zezo357:
will wait for ci then merge, just to be sure there is nothing i missed.

zezo357:
Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.

i cant merge btw :) will leave it to you then

alexmercerind:
You must now have the permission!

zezo357:
> You must now have the permission!

yes i do now, thank you

also builds passed so i am merging

Issue #394, opened 2023-08-22
[Bug Report] Subtitles not showing m3u8


kuki-kuk:
I have a problem to show subtitles when  streaming video using m3u8 link. 
Example:
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-TARGETDURATION:13
#EXT-X-MEDIA-SEQUENCE:808
#EXTINF:12.012000,
25011126_808.ts
#EXTINF:12.012000,
25011126_809.ts
#EXTINF:10.643967,
25011126_810.ts
#EXTINF:13.380033,
25011126_811.ts
#EXTINF:12.012000,
25011126_812.ts

tested on Windows and Android.

Thanks,

femalemonkeyman:
Probably need a few more details but looks like the subtitles are external?

alexmercerind:
Yeah, I don't think this information is sufficient, you can load external subtitles (how it's mostly done on the web):

https://github.com/media-kit/media-kit#load-external-subtitle-track

lKinderBueno:
I have the same issue using embedded subtitles.
This is the video with embedded subtitles: [https://supertest.netsons.org/video_sub.mp4](https://supertest.netsons.org/video_sub.mp4)
This is the srt file embedded in the video [https://supertest.netsons.org/sub.srt](https://supertest.netsons.org/sub.srt)

If I try to set the subtitle from the url, they works. 
`player.setSubtitleTrack(SubtitleTrack.uri('https://supertest.netsons.org/sub.srt'));`

Instead, if I try to load the embedded subtitle, the player doesn't show any text:
`player.setSubtitleTrack(player.state.tracks.subtitle[2]);`

lKinderBueno:
> @lKinderBueno
> 
> One more question, what platform are you on?

I tested it on Android 12 and Windows

alexmercerind:
I've set things to cross-compile for Windows, Android & macOS. I will publish an update in few hours or tomorrow.

alexmercerind:
FYI: This is due to a missing subtitle decoder _i.e._ issue shall not be present on GNU/Linux.

alexmercerind:
A new version has been published, [please refer to the "Installation" section in the README](https://github.com/media-kit/media-kit#installation).

lKinderBueno:
There are two small issues:

- By default it loads the embedded subtitle
- Once the subtitles are enabled, they can't be disabled with ```dart player.setSubtitleTrack(SubtitleTrack.no())```

[...]

Issue #356, opened 2023-08-12
How to listen playing state changed ?


pubpy2015:
Hi,
Currently, I'm playing an rtsp video stream and getting the play status by listening for events:

player.stream.error.listen((error) => debugPrint("error: " + error));
player.stream.buffering.listen((buffering) => debugPrint("buffering: " + buffering.toString()));
player.stream.playing.listen((playing) => debugPrint("playing: " + playing.toString()));
player.stream.log.listen((event) => debugPrint(" mpv: ${event.prefix}: ${event.level}: ${event.text}"));

with available stream, the status sequence is as follows:
playing: false => playing: true => buffering: true => buffering: false

with not available stream (disconnect):
playing: false => playing: true => buffering: true

the error event never fires and buffering still true although connection timeout.
mpv log:
mpv: ffmpeg/demuxer: error: rtsp: method DESCRIBE failed: 404 Stream Not Found
mpv: lavf: error: avformat_open_input() failed
cplayer: error: Failed to recognize file format.

Can I only listen for 1 event where the player's state is updated ?
Here is the set of player's state in VLC
enum PlayingState {
  initializing,
  initialized,
  stopped,
  paused,
  ended,
  buffering,
  playing,
  recording,
  error,
}

alexmercerind:
The error will correctly fire for this situation in the next update.

pubpy2015:
Thanks,

in case of source switching from one stream to another, can you discard the last frame of the previous stream (go to black screen) before connecting to the new stream?

alexmercerind:
In most cases preserving last frame is ideal, I'll see.

pubpy2015:
Hi, my case is:

I use media kit to display video from some network cameras (rtsp stream). When switching from one camera to another, I want to hide the image of the previous camera to show that I am connecting to another camera.

pubpy2015:
Another question: 

By default the player decodes on GPU, but on my 1 old device, with H265 video stream, the video is not displayed. However if I set enableHardwareAcceleration=false, the video stream can display normally. In cases like this, can the media kit automatically switch back to CPU decoding when the GPU fails?

alexmercerind:
I simply do not care.

alexmercerind:
> Hi, my case is:
> 
> I use media kit to display video from some network cameras (rtsp stream). When switching from one camera to another, I want to hide the image of the previous camera to show that I am connecting to another camera.

Raised #418.

PR #418, opened 2023-08-27
fix: reset `Player` state/stream upon `Player.open`; fix: hide last video's frame upon `Player.open`

alexmercerind:
The behavior of `Player` is _mostly_ still the same.
Only 2 changes can be seen (which I believe are favorable) in tests, as it can be seen:
- `buffering` will be restored back to `false` once `Player.open` is called.
- `tracks` will be cleared once `Player.open` is called.

Primarily this is done to reset previously playing `Media`'s state (while newly opened buffering for example) e.g.
- [Last frame stuck.](https://github.com/media-kit/media-kit/issues/356#issuecomment-1675757193)
- [~~Last subtitles stuck.~~ A different fix & test will be needed.](https://github.com/media-kit/media-kit/issues/394#issuecomment-1694491967)

<hr>

I was already familiar with the issues & it'll be cool to have these resolved.

### DIFF
--- a/media_kit/CHANGELOG.md
+++ b/media_kit/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.1.7
 
-fix: close `PlatformPlayer.playlistModeController`
+- fix: close `PlatformPlayer.playlistModeController`
 
 ## 1.1.6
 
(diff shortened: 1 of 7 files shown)


birros (approved):
LGTM!

AALAMILLO:
Thanks a lot for have this in your mind. When we get the next realease? 😃

PR #441, opened 2023-09-03
[Windows] Fix fullscreen for non-primary monitors

### LINKED ISSUE #439: [Bug Report][Win11] When using fullscreen with the window on a second monitor the display is not as expected
## Describe your issue
When you fullscreen the app with the window on a second monitor the window either enlarges or is reduced to 1px width (i believe)

If the second monitor is on the left side in virtual space, the window is reduced to 1px width (second monitor highlighted in blue)
image

If the second monitor is on the right side in virtual space, the ?window spans to the width of both monitors? (second monitor highlighted in blue) 
image

## Reproduction
1. Open a Player
2. Drag the window to your second monitor (whichever is not your primary display on windows)
3. fullscreen

## Expected Behavior
The window Fullscreens and is the width/height of whatever display it is on.

## Screenshots taken from win + tab

Second monitor is on the right side in virtual space
image

Second monitor is on the left side in virtual space
image

Normal fullscreen on primary monitor
image

NotAgentBilly:
Closes #439 

This change corrects the width and height parameters passed to the SetWindowPos function

Previously, the right and bottom coordinates of the monitor were incorrectly used as the window size. Now, the width and height are correctly calculated as the differences between the right and left coordinates and the bottom and top coordinates.

### DIFF
--- a/media_kit_video/windows/utils.cc
+++ b/media_kit_video/windows/utils.cc
@@ -38,8 +38,8 @@ void Utils::EnterNativeFullscreen(HWND window) {
                      &monitor);
     ::SetWindowLongPtr(window, GWL_STYLE, style & ~WS_OVERLAPPEDWINDOW);
     ::SetWindowPos(window, HWND_TOP, monitor.rcMonitor.left,
-                   monitor.rcMonitor.top, monitor.rcMonitor.right,
-                   monitor.rcMonitor.bottom,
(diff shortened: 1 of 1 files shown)


alexmercerind (approved):
LGTM!

Thanks for the pull-request. I don't have a dual-monitor setup, so I didn't face it.

The fix is indeed logical, I made a mistake.

PR #458, opened 2023-09-06
fix: small player.stop leak

23doors:


alexmercerind (approved):
Thanks a lot for noticing!

PR #474, opened 2023-09-12
feat(video-controls): add 'backdropColor' parameter to MaterialVideoControlsThemeData to configure backdrop color that comes up with controls

ashilkn:
Uses the default colour if parameter is not passed.


alexmercerind (approved):
LGTM!

Thanks for the pull-request.

PR #475, opened 2023-09-12
fix: High gpu usage during idle

### LINKED ISSUE #472: [Bug Report][Windows] GPU usage in idle state
os: windows 11

- There is a constant 8% gpu usage in the state where  the video is not even starting to play
- There is a constant 15% gpu usage when the video is finished playing or stopped or paused

This could causes the laptop to drain the battery too quickly

NotAgentBilly:
Sets the state of the controls CircularProgressIndicator to a constant value when we're not buffering.

Fixes #472

### DIFF
--- a/media_kit_video/lib/media_kit_video_controls/src/controls/material.dart
+++ b/media_kit_video/lib/media_kit_video_controls/src/controls/material.dart
@@ -942,10 +942,18 @@ class _MaterialVideoControlsState extends State<_MaterialVideoControls> {
                             child: _theme(context)
                                     .bufferingIndicatorBuilder
                                     ?.call(context) ??
-                                const Center(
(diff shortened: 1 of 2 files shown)


alexmercerind:
Thanks a lot!

I only checked the native code, even tried removing the render loop entirely. The GPU usage in idle state was still apparently.

As far as it seems, [`CircularProgressIndicator` is quite heavy on CPU/GPU](https://github.com/flutter/flutter/issues/13203). This is definitely taking additional resources besides the actual video playback. @birros internally did notify me about this issue few months ago, however it slipped off my mind.

alexmercerind:
It might be wiser to replace `CircularProgressIndicator` entirely with some less-expensive progress indicator.

Issue #480, opened 2023-09-15
[Bug Report] Crash of `mpv_render_context_free` when `Player.open` fails or if not called before `Player.dispose`


birros:
I'm experiencing this crash on macOS, but it may also exist on other platforms:

```
VM Region Info: 0x13e0 is not in any region.  Bytes before following region: 4305701920
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
--->  
      __TEXT                      100a3e000-100a46000    [   32K] r-x/r-x SM=COW  ...cOS/Foo

Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libGL.dylib                   	    0x7ffa1c27978b glDeleteQueries + 18
1   Mpv                           	       0x1022d0ad0 gl_timer_destroy + 32
2   Mpv                           	       0x102299cde timer_pool_destroy + 30
3   Mpv                           	       0x10229ffea gl_video_uninit + 90
4   Mpv                           	       0x10229318c destroy + 28
5   Mpv                           	       0x1022ae620 mpv_render_context_free + 336
6   media_kit_video               	       0x101c39580 TextureHW.disposeMPV() + 34 (TextureHW.swift:105) [inlined]
7   media_kit_video               	       0x101c39580 TextureHW.__deallocating_deinit + 48 (TextureHW.swift:36)
8   media_kit_video               	       0x101c39605 @objc TextureHW.__deallocating_deinit + 21
```

This happens when `Player.dispose` is called after one of these two contexts:
- If `Player.open` is not called
- If `Player.open` is called, but the URL does not exist: `player.open(Media('http://127.0.0.1:1234/foobar.mkv')`

I can reproduce this issue in pure native code, i.e. without Flutter.
If `mpv_command("loadfile", ...)` fails or is not called `mpv_render_context_free` crashes.

PR #481, opened 2023-09-15
fix(video): fix mpv_render_context_free call

birros:
- fix `mpv_render_context_free` crash (#480)

### DIFF
--- a/media_kit_video/common/darwin/Classes/plugin/TextureSW.swift
+++ b/media_kit_video/common/darwin/Classes/plugin/TextureSW.swift
@@ -69,6 +69,7 @@ public class TextureSW: NSObject, FlutterTexture, ResizableTextureProtocol {
   }
 
   private func disposeMPV() {
+    mpv_render_context_set_update_callback(renderContext, nil, nil)
     mpv_render_context_free(renderContext)
   }
 
--- a/media_kit_video/ios/Classes/plugin/TextureHW.swift
+++ b/media_kit_video/ios/Classes/plugin/TextureHW.swift
@@ -98,10 +98,10 @@ public class TextureHW: NSObject, FlutterTexture, ResizableTextureProtocol {
   }
 
   private func disposeMPV() {
-    CGLSetCurrentContext(context)
+    EAGLContext.setCurrent(context)
     defer {
-      OpenGLHelpers.checkError("disposeMPV")
-      CGLSetCurrentContext(nil)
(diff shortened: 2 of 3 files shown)


birros:
@alexmercerind (or someone else who has time) I think this patch should also be applied to Windows and Linux, even if there are no crashes today, it will make the code future-proof.

alexmercerind:
Interesting. I haven't tested the exact situation mentioned in the issue. I will take a look to confirm. I just need to call `mpv_render_context_set_update_callback` with `NULL` at destruction to remove existing render update callback?

**Windows:**

`mpv_render_context_free` is called once all previous render-callbacks are completely executed.

https://github.com/media-kit/media-kit/blob/8926ede69cb11aca45e9ea2b26428202e82ea5de/media_kit_video/windows/video_output.cc#L142
https://github.com/media-kit/media-kit/blob/8926ede69cb11aca45e9ea2b26428202e82ea5de/media_kit_video/windows/video_output.cc#L145-L147

Also: as soon as destructor is hit, a flag ensures that no existing attributes etc. are referenced in future.

https://github.com/media-kit/media-kit/blob/8926ede69cb11aca45e9ea2b26428202e82ea5de/media_kit_video/windows/video_output.cc#L111

**GNU/Linux:**

Only a flag. I'm still confused about threading model.

https://github.com/media-kit/media-kit/blob/8926ede69cb11aca45e9ea2b26428202e82ea5de/media_kit_video/linux/video_output.cc#L40

birros:
> I just need to call `mpv_render_context_set_update_callback` with `NULL` at destruction to remove existing render update callback?

I didn't take the time to check which of the GL context lock or `mpv_render_context_set_update_callback` with `NULL` fixes this issue, I just copied what IINA [does](https://github.com/iina/iina/blob/d7039f3e3f12cd5e6e3fbaea7618d0606f771637/iina/MPVController.swift#L506-L512):

```swift
func mpvUninitRendering() {
  guard let mpvRenderContext = mpvRenderContext else { return }
  lockAndSetOpenGLContext()
  defer { unlockOpenGLContext() }
  mpv_render_context_set_update_callback(mpvRenderContext, nil, nil)
  mpv_render_context_free(mpvRenderContext)
}
```

But I think it's the GL context lock that fixes the issue, as the crash happens when `glDeleteQueries` is called.

Again, I don't know if this crash can exist on other platforms, maybe it's specific to macOS / iOS.

Issue #14, opened 2023-01-06
additional arguments (lib)mpv for RTSP-Stream


filly82:
My Local RTSP-Stream (TV-Stream via Router) needs an additional argument for mpv to work `--rtsp-transport=udp`

cmd: `mpv.exe "rtsp://..." --rtsp-transport=udp`

is this somehow currently possible or will be in future?

Quackdoc:
might be possible to simply pass 

```
mpv_set_option_string(handle_, "rtsp-transport", "udp");
```

but I feel like MPV should implicitly handle this. I think you can specify UDP via URI can't you?

alexmercerind:
Hi @filly82,

I did expose an API for setting libmpv properties/options from Dart.

```dart
final player = Player();
// Check type (only true for Windows & Linux).
if (player?.platform is libmpvPlayer) {
  await (player?.platform as libmpvPlayer?)?.setProperty("rtsp-transport", "udp");
}
```

In theory [this will work](https://github.com/harmonoid/harmonoid/blob/9ccc4ee116b5e4f19c462756403d5db9eb2def6f/lib/core/playback.dart#L324-L332). If it doesn't, sure let us know.

PR #525, opened 2023-10-09
Add profile and options for native MPV player

DipDevStudio:
When working with the camera and viewing the RTSP stream it is required to playback video without delay - for example, during PTZ control. This requires calling the native low-latency profile and setting the untimed flag.

This solution will help to customize and use the player more flexibly.

For example:

`final player = Player(configuration: const PlayerConfiguration(bufferSize: 1 * 1024 * 1024, profile: MPVProfile.lowLatency, options: {'untimed':'yes'}));`

### DIFF
--- a/media_kit/lib/src/player/native/player/real.dart
+++ b/media_kit/lib/src/player/native/player/real.dart
@@ -2367,8 +2367,7 @@ class NativePlayer extends PlatformPlayer {
         },
       );
 
-      properties
-          .addEntries((configuration.options ?? <String, String>{}).entries);
+      properties.addAll(configuration.options ?? <String, String>{});
 
       if (test) {
         properties['vo'] = 'null';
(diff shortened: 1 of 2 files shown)


alexmercerind:
Hi @DipDevStudio,

First of all, I want to appreciate your pull-request to the project.

I have to reject this one, since there's an existing API to set custom libmpv properties/options: https://github.com/media-kit/media-kit/issues/14#issuecomment-1373205132.

[It's intentionally hidden because 1. It's not available for every platform e.g. web 2. It is backend specific e.g. libmpv]

As a project, we only provide support to the API exposed (classes & methods) by package:media_kit e.g. `Player`, `VideoController` etc. Modifying internal backend specific properties/options (libmpv in this case) can cause a number of issues (if not done carefully). It's like forking the package & adding separate code.

In other cases, it has been seen that changing profile can cause issues on Android. Developers are obviously free to do so with the existing API, but not something we can support in a stable way (really big concern).

Thanks!

PR #533, opened 2023-10-14
feat: `VideoState.update` & `VideoViewParameters`

zezo357:
an idea on how to update the fit and other properties of the video widget (sadly i wasnt able to make the notifiers shared between the video widget and the full screen one)

so i will leave this as a draft hopefully someone got an idea on to make them shared without making api changes (this is my limiting point )

---
Edit:

I was able to make them shared. It may not be the cleanest workaround, so if anyone has a cleaner fix, feel free to share.

**How to update the properties:**

Example:
```dart
videoStateKey.currentState?.update(fit: BoxFit.contain);
```

**Public Function Implementation:**
```dart
void update({
  BoxFit? fit,
  Color? fill,
  Alignment? alignment,
  double? aspectRatio,
}) {
  if (fit != null) _fitNotifier.value = fit;
  if (fill != null) _fillNotifier.value = fill;
  if (alignment != null) _alignmentNotifier.value = alignment;
  if (aspectRatio != null) _aspectRatioNotifier.value = aspectRatio;
}
```


alexmercerind:
The idea is indeed very neat.

I just had one minimal thought, how about we compose another data/model class e.g. `VideoStateParameters` (bad-naming feel-free; which implements/overrides `==` & `hashCode`) w/ all these parameters (some exceptions may apply, I leave that upto your requirement):

https://github.com/media-kit/media-kit/blob/b4a851650602c9b55a6bfb5412b28655f34b65aa/media_kit_video/lib/src/video/video_texture.dart#L117-L130

This will allow us to have a single `ValueNotifier` for all the attributes. I don't think there'll be any disadvantage, since we expect the widget to be redrawn upon the `update` method you added. This will allow us to easily add more parameters in future if required & make implementation simpler too.

Thanks!

zezo357:
> The idea is indeed very neat.
> 
> I just had one minimal thought, how about we compose another data/model class e.g. `VideoStateParameters` (bad-naming feel-free; which implements/overrides `==` & `hashCode`) w/ all these parameters (some exceptions may apply, I leave that upto your requirement):
> 
> https://github.com/media-kit/media-kit/blob/b4a851650602c9b55a6bfb5412b28655f34b65aa/media_kit_video/lib/src/video/video_texture.dart#L117-L130
> 
> This will allow us to have a single `ValueNotifier` for all the attributes. I don't think there'll be any disadvantage, since we expect the widget to be redrawn upon the `update` method you added. This will allow us to easily add more parameters in future if required & make implementation simpler too.
> 
> Thanks!

I really like the idea of having a single notifier since the idea of having multiple notifiers is annoying since they are passed everywhere, 


I will check it tomorrow night, hopefully I come up with something that easy to manage for future updates,

zezo357:
added VideoViewParameters, with all the params to cover most users use cases

zezo357:
for context now the pr is ready from my side all values that can be updated are used and removed wakelock since its not updatable and removed controller from notifier since its not used and may result in some bugs since people may update controller mid playing which will lead to undefined behaviour (not tested but i think it may happen)

alexmercerind (approved):
LGTM!

<hr>

I made few improvements.

PR #545, opened 2023-10-19
Fix web

zezo357:
i see you did fix the compilation problem, but i was doing a fix too, so it may be in another way but will leave it to pick which is better


zezo357:
will close this, open if you feel its needed.

PR #586, opened 2023-11-12
feat: `NativePlayer.command`

alexmercerind:
Now that `setProperty`, `observeProperty` & `unobserveProperty` are part of public API...

Let's add last piece of the puzzle.


birros (approved):
LGTM!

PR #601, opened 2023-11-16
fix(video_player_media_kit): `VideoPlayerValue.size` stuck at `Size.zero` after `VideoEventType.initialized`

### LINKED ISSUE #599: video_player_media_kit，VideoPlayer There is a problem with initialization completion
VideoPlayerController，After the asynchronous call of the initialize method is completed, the obtained width and height are 0. 
There is something wrong with the following code. After the initialization is completed, the value is always 0.

```
streamSubscriptions.add(
        player.stream.videoParams.listen(
          (event) {
            width = event.dw ?? 0;  //  adjusted to  //  event.dw；
            height = event.dh ?? 0; //  adjusted to  //  event.dh ；
            if ((width ?? 0) > 0 && (height ?? 0) > 0) {
              notify();
            }
          },
        ),
      );
      streamSubscriptions.add(
        player.stream.tracks.listen(
          (event) {
            // No video track is available i.e. an audio file.
            if (event.video.length == 2) {
              width = 0;
              height = 0;
              notify();
            }
          },
        ),
      );
```

alexmercerind:
Closes #599.

### DIFF
--- a/video_player_media_kit/CHANGELOG.md
+++ b/video_player_media_kit/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.0.5
+
+- fix: `setVolume` implementation
+- fix: `VideoPlayerValue.size` stuck at `Size.zero` after `VideoEventType.initialized`
+
 ## 1.0.4
 
 - build: bump `media_kit_video`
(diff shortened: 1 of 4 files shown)


zezo357 (approved):
cant test it right now, but LGTM.

i think this was the fix? `if (event.video.length == 2 && event.audio.length > 2)`

longgong:
very good

Issue #506, opened 2023-09-28
Crash problem on Windows


Nik96i:
Hi.
I use media_kit Player in my app. I'm developing on Windows 11 and the app crashes without any error. I also built the app and tested it again, same thing happened. it works fine on other PCs running Windows 11 but it has a problem with my PC.
DirectX and VC Redistributable are also installed.
Thanks.

alexmercerind:
Hi @Nik96i,

We did get one similar bug-report few months ago. They had exact same issue on Windows. They ended up re-installing Windows, which fixed it for them. We couldn't find any cause or sufficient information at the time. We only received one such report, other users have no complaints.

You can try to run our demo application (direct executable available in README) & check if that works. A stack trace will be further helpful.

Thanks!

Nik96i:
@alexmercerind Thanks u for ur reply sir. I can Replace my Windows but it may happen in one of our customer's systems, so it would be awesome if this problem resolves. We Cannot release our app like that. I tested the demo app and it also crashed. If u can provide a simple debug/test app which outputs every log or something, I can test it and maybe it provide some hints.
Here I tested the demo app:
[https://file.io/P1v8muczlE28](url)
Thanks again.

alexmercerind:
Hi @Nik96i,

Does switching to software rendering work? Click the "H/W" button on the top-right of the application to make it "S/W".

Nik96i:
@alexmercerind No. it doesn't work.

Nik96i:
Hi @alexmercerind 

This is my crash log in the Windows event viewer.
I think the program cannot load "libmpv-2.dll" file.

```
Faulting application name: media_kit_test.exe, version: 1.0.0.0, time stamp: 0x6514abf2
Faulting module name: libmpv-2.dll, version: 2.0.0.0, time stamp: 0x650fea09
Exception code: 0xc0000094
Fault offset: 0x0000000000134ec4
Faulting process id: 0x0x166C
Faulting application start time: 0x0x1D9F374589E9231
Faulting application path: C:\Users\nik96i\Downloads\media_kit_test_win32_x64\buildwindowsrunnerRelease\media_kit_test.exe
Faulting module path: C:\Users\nik96i\Downloads\media_kit_test_win32_x64\buildwindowsrunnerRelease\libmpv-2.dll
Report Id: e203d615-2c4e-4c6a-a0b8-836902a09c1b
Faulting package full name: 
Faulting package-relative application ID: 
```

[...]

PR #619, opened 2023-11-25
Update native_library.dart

punxch:
Fix errror of loop finding mpv dlls.  

Loop finding dlls will cause exception on Windows platform even though libmpv-2.dll has being founded.  

The fixed exception is:
Exception has occurred.
ArgumentError (Invalid argument(s): Failed to load dynamic library 'mpv-2.dll': error code 126)

### DIFF
--- a/media_kit/lib/src/player/native/core/native_library.dart
+++ b/media_kit/lib/src/player/native/core/native_library.dart
@@ -71,6 +71,7 @@ abstract class NativeLibrary {
         try {
           DynamicLibrary.open(name);
           _resolved = name;
+          return;
         } catch (_) {}
       }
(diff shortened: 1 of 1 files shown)


punxch (commented):
Fix errror of loop finding mpv dlls.  

Loop finding dlls will cause exception on Windows platform even though libmpv-2.dll has being founded.  

The fixed exception is:
Exception has occurred.
ArgumentError (Invalid argument(s): Failed to load dynamic library 'mpv-2.dll': error code 126)

alexmercerind (approved):
LGTM!

alexmercerind:
Thanks a lot for pointing this out. I think this will also resolve #506.

PR #637, opened 2023-12-11
fix: (maybe) Adjust cmake options to have them not try to  define pthread

femalemonkeyman:
This should fix issue 595 about building on fedora. It no longer tries to add the definitions that were coming from cflags :/


birros (approved):
Great work!
I can confirm that the compilation works on fedora 39.

PR #640, opened 2023-12-15
fixed stuck double tap

zezo357:
### Issue Description
- **Problem:** The seek buttons in the video controls were incorrectly remaining hidden while still registering clicks. 
- **Cause:** This behavior was due to both `_hideSeekForwardButton` and `_mountSeekForwardButton` being set to `true` at the same time. The root cause of this issue was traced back to the `TweenAnimationBuilder`'s end value being set to zero which prevents onend from being called.

### Proposed Solution
- **Fix Implementation:** The proposed fix involves a minor yet critical adjustment to the `TweenAnimationBuilder`.
- **Details:**
    - Change the end value from `0.0` to a minimal non-zero value (`0.000001`).
    - This adjustment allows the 'on end' callback to be called to hide and unmount the UI elements.


zezo357:
used animated opacity and timers instead, since the bug still happens and not nothing resets it. :(

PR #641, opened 2023-12-15
[darwin] Run all engine callbacks on the main queue

### LINKED ISSUE #590: [Warning][Windows] `flutter::MethodChannel::InvokeMethod` background thread usage
[ERROR:flutter/shell/common/shell.cc(1004)] The 'com.alexmercerind/media_kit_video' channel sent a message from native to Flutter on a non-platform thread. Platform channel messages must be sent on the platform thread. Failure to do so may result in data loss or crashes, and must be fixed in the plugin or application code creating that channel.

moffatman:
Not just `registry.register` must run on the main thread, but also `registry.unregisterTexture`, `registry.textureFrameAvailable`, and the `MethodChannel` within `textureUpdateCallback`.

I used synchronous block where possible as it makes the memory model lot more clear, and we don't really want or need this stuff to start queueing up in the background.

This will actually fix #590 on iOS, that warning from flutter is about channel usage, not texture registry.

This stuff should be prioritized to change on any other affected platforms if possible, it causes some real crashes

image

### DIFF
--- a/media_kit_video/common/darwin/Classes/plugin/VideoOutput.swift
+++ b/media_kit_video/common/darwin/Classes/plugin/VideoOutput.swift
@@ -11,6 +11,7 @@
 // To improve the user experience, a worker is used to execute heavy tasks on a
 // dedicated thread.
 public class VideoOutput: NSObject {
+  // Will be called on the main thread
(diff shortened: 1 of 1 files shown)


birros (approved):
Thx for the fix, I'm quite busy at the moment (moving).

Issue #645, opened 2023-12-16
delete media_kit.jar when verification failed


Passer-by:
I found that when the download fails, there is also a default-arm64-v8a.jar file, but the size of this file is 0 kb. According to the judgment in [build.gradle](https://github.com/media-kit/media-kit/blob/main/libs/android/media_kit_libs_android_video/android/build.gradle#L101-L112) it is considered as having a file and will undergo an MD5 verification.

Passer-by:
Perhaps it can be directly deleted after the verification fails. After all, this is just a cache file.

josh4500:
This is an annoying issue and still can't get past this

khomin:
Still not working

```
Build file 'user/.pub-cache/hosted/pub.dev/media_kit_libs_android_video-1.3.6/android/build.gradle' line: 102

* What went wrong:
A problem occurred evaluating project ':media_kit_libs_android_video'.
> media_kit: arm64-v8a JAR verification failed.
```

zezo357:
https://github.com/media-kit/media-kit/pull/656 this pr fixes it, but sadly its not merged yet

PR #656, opened 2023-12-25
fix: new gradle logic for downloading

abdelaziz-mahdy:
improving android gradle to be like windows logic, fixing https://github.com/media-kit/media-kit/issues/645


abdelaziz-mahdy:
Check your internet connection, the ci works so it's not a problem with the code, also you can rerun without flutter clean and the code will redownload the broken files

alexmercerind (approved):
Merging. I will check if it breaks.

PR #661, opened 2023-12-30
libass fix not  displayed on Android

user1121114685:
It took me a long time to find the reason why srt and ass subtitles are occasionally displayed and sometimes not displayed. This update allows subtitles to be loaded in the default font. If more fonts are used, fonts.conf is required.

Reference link
https://mpv.io/manual/master/#options-sub-fonts-dir


user1121114685:
PixPin_2023-12-30_22-07-29
libassAndroidFontFamilyName

PR #647, opened 2023-12-17
Combined Fixes from PR #644, #640, and #618

abdelaziz-mahdy:
This pull request combines changes from several individual pull requests to avoid merge conflicts. The included pull requests are as follows:

- PR [#644](https://github.com/media-kit/media-kit/pull/644)
- PR [#640](https://github.com/media-kit/media-kit/pull/640)
- PR [#618](https://github.com/media-kit/media-kit/pull/618)

feel free to merge this incase all is good

summary 

**Overview:**
This pull request introduces a series of enhancements and bug fixes to the video player controls, improving the overall user experience and responsiveness. Key areas of focus include widget width handling, gesture detection, and the double-tap seek UI, along with a new feature for mouse visibility control.

**Fixes and Enhancements:**

1. **Widget Width Handling Issue:**
   - **Problem:** Inaccurate control scaling due to using MediaQuery for deriving control widths, causing issues in non-full-screen modes.
   - **Fix:** Implemented LayoutBuilder for dynamic width acquisition, ensuring correct sizing and positioning of controls in various player sizes.

2. **Gesture Detection Delay:**
   - **Problem:** Delay in gesture detection impacting immediate control responsiveness.
   - **Fix:** Segmented logic and immediate touch event listener akin to YouTube's approach for faster control display on touch.

3. **Double Tap Seek UI Adjustment:**
   - **Problem:** Overlapping of double-tap-to-seek function with normal interaction areas.
   - **Fix:** Segmented layout (2,1,2 ratio) with double-tap-to-seek on sides and regular taps in the center, mirroring YouTube's UI for clear differentiation and intuitive use. can be edited by developers

4. **Seek Button Visibility Issue:**
   - **Problem:** Seek buttons hidden yet clickable due to conflicting _hideSeekForwardButton and _mountSeekForwardButton states.
   - **Fix:** implemented animated opacity with timers for consistent behavior.

**New Feature:**

- **Hide Mouse on Controls Removal:**
  - **Addition:** New `hideMouseOnControlsRemoval` property in `MaterialDesktopVideoControlsThemeData`.
  - **Functionality:** Hides mouse cursor when video controls are removed, offering a cleaner full-screen video experience.
  - **Default Behavior:** Set to false to maintain backward compatibility.
  - **Windows:** https://github.com/media-kit/media-kit/pull/618#issuecomment-1826046719
**Usage Example:**

```dart
MaterialDesktopVideoControlsThemeData(
  // ... other properties ...
[...]

PR #742, opened 2024-03-22
Adding single tap for play/pause

sapphire008:
Modifying `_MaterialDesktopVideoControlsState` class of `material_desktop.dart` to allow single tap to toggle play/pause.
Implemented method `onTapDown` under `GestureDetector`.
Note that it may not be a good solution to wrap another `GestureDetector` around the main controller since there will be a significant delay (tried that, user will see a delay before the video is actually paused).
This should mimic YouTube's behavior here: single tab will always trigger play pause, but this will also slightly impact double tapping to go into full screen, where the video will be pause briefly. Region below the bottom navigation bar is protected from single-tap when it is shown to prevent play / pause.


abdelaziz-mahdy:
merged in https://github.com/media-kit/media-kit/pull/647

PR #748, opened 2024-03-26
fix: Use a Scaffold as the outermost widget on fullscreen video pages

refi64:
Using a plain Material means that functionality that's tied to the nearest Scaffold, suck as SnackBars, won't actually be displayed on the page.


alexmercerind (approved):
LGTM!

alexmercerind:
This breaks the fullscreen. Reverting...

PR #749, opened 2024-03-26
feat: Add support for including libass subtitles in screenshots

refi64:
On the native backend, this is essentially available for free: we just need to swap the flag given to the `screenshot-raw` command.

This option is simply ignored by the web backend.


alexmercerind (approved):
LGTM!

I have few nits, I will resolve myself.

PR #766, opened 2024-04-09
chore: fix ci for android and added caching to improve speed

abdelaziz-mahdy:
fix ci for android by setup java 

use flutter action v2 and use its caching to improve speed


alexmercerind (approved):
LGTM!

PR #780, opened 2024-04-21
Add support for milliseconds in `start` and `end` for `Media`

default-anton:
mpv allows the use of `[+|-][[hh:]mm:]ss[.ms]` for `start` and `end`. Currently, the implementation drops the `ms` part.

This PR adds support for milliseconds in `start` and `end` for `Media`.


abdelaziz-mahdy:
pull from main to fix the ci that fails.

PR #786, opened 2024-04-23
chore: set cache to yes

abdelaziz-mahdy:
setting cache to yes instead of auto, since in some cases the cache were not working on some sources 

it can be set in users apps, but the default i think should be with cache enabled on all sources


alexmercerind (approved):
NIT: Use single quotes.

LGTM!

Issue #755, opened 2024-03-30
 Display Buffering Percentage


fsmdeveloper:
Is there any way to show the buffering percentage on a video? For example, displaying 'Video is loading... 10%... 50%... 99%... and then play.

isaacy13:
Hey @fsmdeveloper, yup it's possible

Check out Player.stream.position -- you can listen to changes and update your UI's state accordingly.

![image](https://github.com/media-kit/media-kit/assets/65752735/ec702737-c4b9-420a-8e2b-a37e4848bcc8)

fsmdeveloper:
[isaacy13](https://github.com/isaacy13) 

` player.stream.position.listen((currentDuration) {
        print(currentDuration);
      });`
      
this only gives me the current duration. How can I show the buffering percentage? If I seek the video, the buffer percentage will be displayed after reaching 100%, and the video will play. If I seek again, it will display the buffering percentage again. I hope you understand what I want to implement

zezo357:
I think you can figure out an approximation of the buffering using the current positions and the currently buffered one. For example, if the cache usually plays after the difference is 5 seconds, you could make the percentage based on that. Note: This is a workaround, not an official fix nor ideal.

![Buffering Example](https://github.com/media-kit/media-kit/assets/25157308/72114c6d-5f72-4069-981b-1e909ee9d7d6)

`player.stream.buffer.listen((event) {});`

isaacy13:
My bad, misread post 😃

fsmdeveloper:
> I think you can figure out an approximation of the buffering using the current positions and the currently buffered one. For example, if the cache usually plays after the difference is 5 seconds, you could make the percentage based on that. Note: This is a workaround, not an official fix nor ideal.
> 
[...]

PR #789, opened 2024-04-25
chore: adding buffering Percentage state and streams

abdelaziz-mahdy:
fixing https://github.com/media-kit/media-kit/issues/755 

using the property of cache-buffering-state which returns the percentage of the buffering


alexmercerind (approved):
LGTM!

PR #791, opened 2024-04-27
improving seek speed

abdelaziz-mahdy:
image

we were using `absolute` which uses exact and from the docs looks like that is slow, when testing with the new code it seems faster, but i dont feel that the workaround i did to allow such logic is good, 

so this needs further testing and a better/cleaner position logic

# issues i found which needs to be fixed before merging:
- [x] multiple seeks in the same time is not working correctly due to position not being updated 

from my testing for now it should be all good and seeking is faster when called multiple times (that was a problem in general ) 
and seeking is 50% faster from what i feel

### DIFF
--- a/media_kit/lib/src/player/native/player/real.dart
+++ b/media_kit/lib/src/player/native/player/real.dart
@@ -765,6 +765,10 @@ class NativePlayer extends PlatformPlayer {
     }
 
     if (synchronized) {
+      if (duration == lastSeekPosition) {
+        return Future.value();
+      }
+      lastSeekPosition = duration;
       return lock.synchronized(function);
     } else {
       return function();
@@ -2696,6 +2700,10 @@ class NativePlayer extends PlatformPlayer {
   ///
   /// This is used to prevent [state.buffering] being set to `true` when [pause] or [playOrPause] is called.
   bool isBufferingStateChangeAllowed = true;
+
+  /// last seek position
+  /// this is used to prevent seeking to the same position twice
+  Duration? lastSeekPosition;
 
   /// Current loaded [Media] queue.
   List<Media> current = <Media>[];
--- a/media_kit/lib/src/player/native/player/real.dart
+++ b/media_kit/lib/src/player/native/player/real.dart
@@ -753,11 +753,7 @@ class NativePlayer extends PlatformPlayer {
 
       await compute(
         _seek,
-        _SeekData(
-          ctx.address,
-          NativeLibrary.path,
-          duration,
-        ),
+        _SeekData(ctx.address, NativeLibrary.path, duration, state.position),
       );
 
       // It is self explanatory that PlayerState.completed & PlayerStream.completed must enter the false state if seek is called. Typically after EOF.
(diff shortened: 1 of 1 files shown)


abdelaziz-mahdy:
i cant figure out why the test case is failing since the seek works, but looks like its not as accurate as absolute in exchange for speed i think thats a good trade.

user1121114685:
Precise jump positioning is really not necessary. Users don’t click very accurately. I agree with this update.


In addition, I have a question that is not related to this update. I would like to ask about whether the automatic next episode is triggered by mpv? Or media-kit. I didn't find the relevant code for the next episode in the media-kit.

abdelaziz-mahdy:
> Precise jump positioning is really not necessary. Users don’t click very accurately. I agree with this update.
> 
> In addition, I have a question that is not related to this update. I would like to ask about whether the automatic next episode is triggered by mpv? Or media-kit. I didn't find the relevant code for the next episode in the media-kit.

its handled by mpv, and can you test this pr and see if the seek is faster or not in your case

user1121114685:
> > Precise jump positioning is really not necessary. Users don’t click very accurately. I agree with this update.
> > In addition, I have a question that is not related to this update. I would like to ask about whether the automatic next episode is triggered by mpv? Or media-kit. I didn't find the relevant code for the next episode in the media-kit.
> 
> its handled by mpv, and can you test this pr and see if the seek is faster or not in your case

Thanks, I'd like to test it, but one of these days I'll be back in a small town where I don't have a computer to program with. I can test it after I come back, which will take 3 days. Thank you for your enthusiastic reply.

media_kit/lib/src/player/native/player/real.dart
```diff
@@ -2752,10 +2758,11 @@ void _seek(_SeekData data) {
   final mpv = generated.MPV(DynamicLibrary.open(data.lib));
   final ctx = Pointer<generated.mpv_handle>.fromAddress(data.ctx);
   // ---------
-  final duration = data.duration;
+  final duration = data.seekDuration - data.videoPosition;
   // ---------
   final value = duration.inMilliseconds / 1000;
-  final command = 'seek ${value.toStringAsFixed(4)} absolute'.toNativeUtf8();
+
+  final command = 'seek ${value.toStringAsFixed(4)}'.toNativeUtf8();
```
default-anton, line 2765: By default, mpv does `seek relative+keyframes <number>`. The tests are failing because seeking is no longer `absolute`. You can do the following:
```dart
final command = 'seek ${value.toStringAsFixed(4)} absolute+keyframes'.toNativeUtf8();
```

I don't think this should be a default though, but it might be worth adding a new player configuration to either use accurate but expensive or inaccurate but fast seeking method.

abdelaziz-mahdy: I did try this configuration it fails very badly if you seek multiple times.

And the default for mpv from the documentation is `relative` from what I found 


For the configuration part I agree that it can be optional, but I don't have any idea how to expose the option cleanly

alexmercerind:
No. This is not a solution. It's not really ideal to have 2-3 seconds mismatch, many of our users actually rely on accurate seek (which in-fact is default in most applications).

alexmercerind:
Of course... If one wishes to experiment themselves, they are free to use the hidden `setProperty` API.

abdelaziz-mahdy:
from my test the mismatch was close to 0.5 sec but if you got higher values i dont think this can work, so i agree with you

default-anton:
I tried this yesterday. It couldn’t let me seek 15s forward on a long video file. It was jumping back to the closest keyframe.

For certain use cases that’s probably not an issue, so we can at least file an issue to make this configurable

alexmercerind:
> Of course... If one wishes to experiment themselves, they are free to use the hidden `setProperty` API.

^^

PR #803, opened 2024-05-02
Fix deadlock disposing event loop

moffatman:
I found this deadlock when creating and disposing lots of players. I guess it can happen if dispose comes in while handling the forwarding of an event to dart.


alexmercerind (approved):
LGTM!

PR #814, opened 2024-05-11
Only one getVideoOutParams

moffatman:
Just a little optimization, because these calls to `MPVHelpers.getVideoOutParams` are not allowed at this time (in render callback), it causes some jank (at end of looping video is most obvious). Personally I work around it by eagerly using setSize. However, we can also take a simple optimization to avoid calling it twice for both width and height when it returns both.

### DIFF
--- a/media_kit_video/common/darwin/Classes/plugin/VideoOutput.swift
+++ b/media_kit_video/common/darwin/Classes/plugin/VideoOutput.swift
@@ -33,8 +33,7 @@ public class VideoOutput: NSObject {
   private var height: Int64?
   private var texture: ResizableTextureProtocol!
   private var textureId: Int64 = -1
-  private var currentWidth: Int64 = 0
-  private var currentHeight: Int64 = 0
+  private var currentSize: CGSize = CGSize.zero
   private var disposed: Bool = false
 
   init(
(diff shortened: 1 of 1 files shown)


birros (approved):
LGTM!

Indeed, I agree.

media_kit_video/common/darwin/Classes/plugin/VideoOutput.swift
```diff
@@ -185,31 +177,23 @@ public class VideoOutput: NSObject {
...
+                height: Double(height!)
+            )
+        }
+        
+        let params = MPVHelpers.getVideoOutParams(handle)
+        return CGSize(
+            width: Double(width ?? (params.rotate == 0 || params.rotate == 180
+                                    ? params.dh
+                                    : params.dw)),
+            height: Double(height ?? (params.rotate == 0 || params.rotate == 180
```
23doors, line 194: @moffatman seems like a mistake?
Previously: width = params.rotate == 0 || params.rotate == 180 ? dw : dh
Now it's: width = params.rotate == 0 || params.rotate == 180 ? dh : dw (swapped)
Same for height.

moffatman: whoops https://github.com/media-kit/media-kit/pull/814

PR #815, opened 2024-05-11
Asynchronous setProperty / command

moffatman:
This removes all noticeable blocking of dart/ui thread

Some notes/questions for @alexmercerind 

1. I replaced the "play() -> get_property -> playOrPause" pattern with just set_property('pause', false). I got bad race conditions with the old way. Not sure why you changed to it originally?
2. I can move `seek` back to the main thread now :)
3. I didn't change screenshot() since it has one-off use of command-with-result. But also it uses `package:image` to do some conversion in dart-space, so other Isolate is still a good idea.
4. I didn't implement getProperty. Because the memory ownership there gets a little more complicated. But after rebasing my change here, I see there is now NativePlayer.getProperty. Can think about it some more...

### DIFF
--- a/media_kit/lib/src/player/native/player/real.dart
+++ b/media_kit/lib/src/player/native/player/real.dart
@@ -172,32 +172,7 @@ class NativePlayer extends PlatformPlayer {
       );
 
       // Enter paused state.
-      {
-        final name = 'pause'.toNativeUtf8();
-        final value = calloc<Uint8>();
-        mpv.mpv_get_property(
-          ctx,
-          name.cast(),
-          generated.mpv_format.MPV_FORMAT_FLAG,
-          value.cast(),
-        );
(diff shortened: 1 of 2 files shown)


moffatman:
I guess one more thing to do as well, apply similar changes to `media_kit_video/lib/src/video_controller/android_video_controller/real.dart`

alexmercerind:
This looks genuinely good to me... I'm gonna take some time. I had added all that logic (even if it looks meaningless or bad at places) after testing compatibility with most mpv versions. I'm gonna try to land this.

moffatman:
i shipped it for a while and works well. the problem with test is because of play/pause changed implementation. some of the playing state checks are wrong during dispose i think

alexmercerind (approved):
Thanks a lot!

Issue #819, opened 2024-05-13
[Linux] GPU Acceleration


oliver-hrkltz:
Hello!

First: Thank you for offering a library which supports all Flutter build targets! 

To increase the performance on our kiosk systems I would like to switch from the web build to the linux build. During my tests I noticed that the video playback takes much more CPU if I build the linux version compared to running the web build. And of course the FHD video is also laggy on the linux build.

So I guess the GPU acceleration is not working. The CPU is a Intel N4120 with Intel UHD Graphics running on Kubuntu 24.04. 

Linux build:
```bash
Cannot load libcuda.so.1
media_kit: VideoOutput: Using H/W rendering.
```

If I play the video directly in from the terminal with mpv it works fine and the libcuda.so.1 message is not printed.

Can I somehow provide additional information for you to validate if the hardware acceleration is active or not?

abdelaziz-mahdy:
you can disable the hardware acceleration and check the performance if better or not, https://github.com/media-kit/media-kit?tab=readme-ov-file#display-the-video

for the web, web uses native browser video playback so its not the same as any other platform under the hood

abdelaziz-mahdy:
did you follow this https://github.com/media-kit/media-kit?tab=readme-ov-file#gnulinux-1

oliver-hrkltz:
Update: Fixed for me by adding `mpv_set_option_string(self->handle, "hwdec", "vaapi")` to video_output.cc. It's still missing the cuda package but the acceleration works now.

As a general fix I guess it could be enough to include/read the mpv config file (/etc/mpv/mpv.conf).

abdelaziz-mahdy:
i dont know much about the configuration of mpv, if that works for you can you add a pr so we see if it works on our ci too?

abdelaziz-mahdy:
i dont know but you can test it using 

```
    final videoController = VideoController(player,
        configuration: VideoControllerConfiguration(hwdec: "vaapi"));
```

@oliver-hrkltz i think it should work for you too instead of changing the video_output.cc

PR #820, opened 2024-05-13
fix: HW acceleration on Linux with Intel GPU

oliver-hrkltz:
Add `--hwdec=VAAPI` parameter to mpv setup to fix the HW acceleration on Linux with Intel GPU.


abdelaziz-mahdy:
check this too https://github.com/media-kit/media-kit/issues/819#issuecomment-2107955762 , if it works then no need to apply it in video_output file

alexmercerind:
It's not a good idea to enforce VAAPI for every single machine out there. The selection of hwdec also depends upon the version of mpv that you're using etc., which can again greatly vary on GNU/Linux, since we use system installation of libmpv.

Issue #776, opened 2024-04-17
memory leaks？


liuchuancong:
After playing the video for an hour in Windows, the content soared from 300MB to 1G. The non-caching attribute has been set. Is this a memory leak in the player?

Here is my simple code :
```
player = Player();
      if (player.platform is NativePlayer) {
        await (player.platform as dynamic).setProperty('cache', 'no');
      }
      mediaPlayerController = media_kit_video.VideoController(player);
      setDataSource(datasource);
      mediaPlayerController.player.stream.playing.listen((bool playing) {
        if (playing) {
          isPlaying.value = true;
        } else {
          isPlaying.value = false;
        }
      });
      mediaPlayerController.player.stream.error.listen((event) {
        if (event.toString().contains('Failed to open')) {
          hasError.value = true;
          isPlaying.value = false;
        }
      });
      mediaPlayerControllerInitialized.value = true;
```

justbendev:
I I'm running 16 RTSP Streams and ram usage go up pretty quick to around **2900 MB** in only 15 minutes.
_Note : 16 Player instances alone without any media is already 700-800MB of memory_

The cache should be default be limited to 32MB but i don't think this is actually being applied.
After seeing you use setProperty i investigated the function and read all available mpv options and applied theses options.

```dart
Player ply = Player();
      (ply.platform as dynamic).setProperty('cache', 'no');                   // --cache=<yes|no|auto>
      (ply.platform as dynamic).setProperty('cache-secs', '0');               // --cache-secs=<seconds> with cache but why not.
      (ply.platform as dynamic).setProperty('demuxer-seekable-cache', 'no');  // --demuxer-seekable-cache=<yes|no|auto> Redundant with cache but why not.
      (ply.platform as dynamic).setProperty('demuxer-max-back-bytes', '0');   // --demuxer-max-back-bytes=<bytesize>
      (ply.platform as dynamic).setProperty('demuxer-donate-buffer', 'no');   // --demuxer-donate-buffer==<yes|no>

```

Now it's stable and won't go over 1,688 MB after hours of running it

_All my memory usage numbers are in Debug mode_

liuchuancong:
@justbendev Thanks for the recovery, I'll try to see if the problem recurs, you're my hero!

liuchuancong:
This problem still exists,increase 30M every 10 minutes when playing m3u8 media files

sherlockvn:
[...]

PR #823, opened 2024-05-15
chore: adding missing dispose

abdelaziz-mahdy:
thanks to https://github.com/media-kit/media-kit/issues/776#issuecomment-2111430301 pointing out that we are missing some disposes, i added them to avoid memory leaks

also i moved the context lookup to `didChangeDependencies` to fix `Looking up a deactivated widget's ancestor is unsafe.`

the dispose is conditional to avoid wrong disposes when moving out of the fullscreen to normal player


moffatman (approved):
LGTM

abdelaziz-mahdy:
The problem was the old context being referenced in that hash map. so will solve the conflict and then merge

PR #825, opened 2024-05-16
do not log plugin registration in stdout

ndusart:
For a flutter desktop application, I need to be able to just return some information on the standard output when some flags are passed from the CLI and exit the application.

All works fine except that media_kit plugin prints a string to the stdout when it is registered.

This PR just removes these prints.


ndusart:
@alexmercerind @moffatman possible to have a look at it ?

moffatman:
It's OK with me. Let's wait for @alexmercerind

abdelaziz-mahdy:
for a workaround you can ignore those messages?

moffatman:
Not a workaround lmao. Let's remove them

alexmercerind:
I'm gonna let them stay... having these is going to confirm that users added those libs packages.

moffatman:
Its just junk in the log imo....

PR #845, opened 2024-05-30
update change logs to prepare release

abdelaziz-mahdy:
this has to be merged first https://github.com/media-kit/media-kit/pull/647#pullrequestreview-2088298625

PR #602, opened 2023-11-17
chore: adding custom controls examples.

abdelaziz-mahdy:
- Added new test files for custom controls:
  - `12.custom_mobile_controls.dart`
  - `13.custom_desktop_controls.dart`
  - `14.custom_adaptive_controls.dart`

PR #851, opened 2024-06-03
Fix specific case of null

abdelaziz-mahdy:
The fallback being cleared from the internal `dispose` method caused the context to sometimes be null. Moving the fallback clearing to the parent widget makes it safe to be cleared without causing this null issue.

This problem did not happen in our examples, but it did occur in the full screen example from [media-kit PR #602](https://github.com/media-kit/media-kit/pull/602). Specifically, the issue arises when trying to exit the video player after entering and exiting full screen mode.

I believe this fix is necessary before merging [media-kit PR #845](https://github.com/media-kit/media-kit/pull/845).


abdelaziz-mahdy:
i was testing on the branch of the examples, let me know if i should separate the fix, i think the examples should be added to avoid this problem happening again.

moffatman (approved):
I don't understand this code and it has a bad smell, whenever you are tracking contexts and doing this math in dispose()..... but anyway will trust you

abdelaziz-mahdy:
> I don't understand this code and it has a bad smell, whenever you are tracking contexts and doing this math in dispose()..... but anyway will trust you

i am trying to find a cleaner way to handle the disposes but i cant find it sadly :_(

PR #870, opened 2024-06-19
removed libmpv.so.1 dependency as libmpv.so.2 is the new thing.

misterhackerman:
libmpv.so.1 is not present on most Linux systems with mpv installed. that issue was remedied by symlinking libmpv.so.2 to libmpv.so.1. this commit only removes that dependency.


alexmercerind:
@misterhackerman 

I don't think there is a problem in maintaining backward compatibility. 

Something you may find interesting:

https://github.com/media-kit/media-kit/commit/143e752650a86501c0eb75844f6c4672aaa219b6

PR #874, opened 2024-06-23
Fixes volume slider affected by click-and-drag

pvogt:
When performing click-and-drag on the seek bar or on the video surface, the code that's commented out affects the volume slider.


alexmercerind:
I'm not sure if it's a good thing to do.

Issue #856, opened 2024-06-06
Can not Find Header When Add to Existing IOS App Using Module Way


Satsrag:
https://github.com/media-kit/media-kit/blob/ff00dae2b9bb6e66cdf15a7ecc1f99d0b7d6e3b4/media_kit_native_event_loop/ios/media_kit_native_event_loop.podspec#L43

Please see this line, it should be replaced with "$(inherited)" "$(PODS_TARGET_SRCROOT)/../common/darwin/Headers".

In my use case: My native iOS project and [Flutter Module](https://docs.flutter.dev/add-to-app/ios/project-setup#option-a-embed-with-cocoapods-and-the-flutter-sdk) are in two different dir. So using `$(PROJECT_DIR)` here can not find header files. There should use `$(PODS_TARGET_SRCROOT)`. 

I only changed this line and it works for me. However, there are multiple places to fix.

abdelaziz-mahdy:
feel free to open a pr with the mentioned fix to be reviewed by birros when he is free

PR #877, opened 2024-06-25
update header dir reference from pod root

Satsrag:
for #856. In my use case: My native iOS project and [Flutter Module](https://docs.flutter.dev/add-to-app/ios/project-setup#option-a-embed-with-cocoapods-and-the-flutter-sdk) are in two different dir. So using `$(PROJECT_DIR)` here can not find header files. There should use `$(PODS_TARGET_SRCROOT)`.


alexmercerind:
Does this work @Satsrag ?

Satsrag:
> Does this work @Satsrag ?

It works for me. But I'm not sure other use case. And I'm also not sure should we use $(PODS_TARGET_SRCROOT) to following two line?
image

alexmercerind:
I didn't write macOS/iOS video module... so don't have any opinions.

Please do push any final changes that you have, let's see if CI breaks.

Satsrag:
I have pushed all my changes.

Satsrag:
This issue seems to have been solved on the latest version. I close this first. If this reoccurs, I open it again.

Issue #357, opened 2023-08-12
App does not build on linux


Wachu985:
**First of all thanks for your work:**
I am using Manjaro KDE to run the application. I have followed all the steps in the README and I have everything installed. I am using version 3.10.6 of Flutter and version 0.36.0 of mpv. Everything is installed correctly and I have compiled other linux apps fine so far. Thanks in advance

### Logs
`Launching lib/main.dart on Linux in debug mode...
main.dart:1
CMake Error: CMake was unable to find a build program corresponding to "Unix Makefiles".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
Exception: Build process failed
Exited`

alexmercerind:
https://github.com/media-kit/media-kit/actions/runs/5799504452/job/15719647940

alexmercerind:
Sorry can't help. Not enough details. Our CI succeeds & I can build locally.

TheOneWithTheBraid:
This issue is caused by the line https://github.com/media-kit/media-kit/blob/main/libs/linux/media_kit_libs_linux/linux/CMakeLists.txt#L93

It looks like the CMAKE related variables set by Flutter for the plugin build are not passed over to the `mimalloc` build unless they are globally declared around.

Yep, GitHub actions do not seem to encounter this issue, but in a plain Debian Docker image I can reproduce it.

TheOneWithTheBraid:
https://github.com/media-kit/media-kit/blob/main/libs/linux/media_kit_libs_linux/linux/CMakeLists.txt#L95

> add_subdirectory() is too much work. Alternatively building it through command line.

I guess here's the response. @alexmercerind I'd kindly ask to reopen this issue since it's now clear how it's caused (and reproducible).

I guess the most meaningful way to shim mimalloc would be simply depending on the system-provided mimalloc for linking ?

- `libmimalloc-dev` is packaged on Debian
- `mimalloc` is packaged on Arch

And likely any other Linux distribution too.

PR #883, opened 2024-06-28
chore: refactor mimalloc import

TheOneWithTheBraid:
Fixes: #357

### DIFF
--- a/README.md
+++ b/README.md
@@ -1673,7 +1673,15 @@ There are other ways to bundle these within your app package e.g. within Snap or
 
 You should consider replacing the default memory allocator with [mimalloc](https://github.com/microsoft/mimalloc) for [avoiding memory leaks](https://github.com/media-kit/media-kit/issues/68).
 
-This is as simple as [adding one line to `linux/CMakeLists.txt`](https://github.com/media-kit/media-kit/blob/d02a97ce70b316207db024401fb99e3f4509a250/media_kit_test/linux/CMakeLists.txt#L92-L94):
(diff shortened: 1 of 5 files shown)


TheOneWithTheBraid:
May I kindly ask for review ?

alexmercerind:
Hi @TheOneWithTheBraid, unfortunately this pull-request is not good.

We want to statically link with mimalloc & not depend upon the one from package manager, as it makes things very much distribution specific.

TheOneWithTheBraid:
@alexmercerind I understand your concern, still, I see a couple of issues with the current implementation :

- you already *do* require `mpv` as distribution dependency - it does not change any effort in setup whether to add one or two dependencies since mimalloc is packaged on seriously any distribution and also easily ships on snap
- if someone does not add mimalloc as dependency, it still works - the plugin can also run without mimalloc - this didn't change
- static linking highly increases my application size and prevents from fast security updates provided by distributions
- the current static linking implementation is awful
  - you call CMAKE in CMAKE - without passing any parameters to it
  - in most build environments, this will likely fail
  
While I understand your concern, I do not think the current approach is a sustainable way to further go.

TheOneWithTheBraid:
@alexmercerind could you elaborate why this PR is closed ? The issue addressed here is still present. I'm open to solve it differently if you prefer static linking, but simply closing a PR without even acknowledging the issue reported exists is a pretty bad attitude.

Another solution could e.g. be using `include_dir` in CMake rather than calling CMake from CMake.

Issue #886, opened 2024-06-30
[media_kit] stub.dart implementations differ from real.dart signature in NativePlayer and WebPlayer


TheOneWithTheBraid:
I figured while debugging builds for web explicitly calling `NativePlayer.setProperty` that this method is not stubbed into the `NativePlayer` stub implementation. Same for most `WebPlayer` methods.

I'd propose to simply resolve this issue by adjusting the stub to be a practical representation of the real implementation - including all of its public members as best practice.

My build logs calling `setProperty` : https://gitlab.com/polycule_client/polycule/-/jobs/7224970988

I will provide you with a corresponding Pull Request in a moment.

PR #887, opened 2024-06-30
fix: adjust stub templates to comply with real types

TheOneWithTheBraid:
This harmonizes the signature of the `stub.dart` and `real.dart` of both `NativePlayer` and `WebPlayer` as best practice for all its public members potentially used by third party packages (in my case [`just_audio_media_kit` bindings](https://pub.dev/packages/just_audio_media_kit)) or directly by application developers.

I see no potential regressions in it.

Fixes: #886


TheOneWithTheBraid:
May I kindly ask for review ?

alexmercerind:
Sounds like a good idea... but we recommend to cast `platform.player` as `dynamic` not `NativePlayer` or `WebPlayer`...

I don't really want to add stub methods... I'm pretty sure some guy will come in future & forget to add stub variant for the other platform... which will cause things to break.

PR #893, opened 2024-07-03
fix: wrong value in brightness builder callback

areille:
`brightnessIndicatorBuilder` is called with `_volumeValue` instead of `_brightnessValue`


alexmercerind (approved):
LGTM!

PR #916, opened 2024-08-08
chore: auto pass changes to fullscreen

abdelaziz-mahdy:
removes the need to call .update, set state or rebuilding the widget in general will pass the changes to the full screen widget


media_kit_video/lib/src/video/video_texture.dart
```diff
@@ -229,6 +229,45 @@ class VideoState extends State<Video> with WidgetsBindingObserver {
     super.didChangeDependencies();
   }
 
+  @override
+  void didUpdateWidget(covariant Video oldWidget) {
```
moffatman, line 233: I think covariant should be removed

media_kit_video/lib/src/video/video_web.dart
```diff
@@ -200,6 +200,45 @@ class VideoState extends State<Video> with WidgetsBindingObserver {
     );
   }
 
+  @override
+  void didUpdateWidget(covariant Video oldWidget) {
```
moffatman, line 204: remove covariant here too

moffatman (approved):
just a minor thing

abdelaziz-mahdy:
done.

PR #932, opened 2024-08-23
Chore  making all state private due to mix of states

abdelaziz-mahdy:
this fixes confusing video state which are used wrongfully, also fixes internal bugs that were using the wrong state too


moffatman:
I don't think it's a good idea to make all the widget fields private, rather not merge this

abdelaziz-mahdy:
but that will make sure no one is using the wrong state? i can fix the internal bugs and expose video view

alexmercerind:
Not really gonna touch anything now...

PR #936, opened 2024-08-26
fix:long press video speed reset issue

user1121114685:
Resolved an issue where long pressing to speed up video playback would incorrectly reset the speed to 1.0x when the user was using a non-default speed (e.g., 1.25x, 1.5x).


alexmercerind:
Is this good to be merged @user1121114685 ?

user1121114685:
Of course, this issue is easy to reproduce and this update fixes it.

PR #946, opened 2024-09-06
Added showControlsWhenPaused

### LINKED ISSUE #940: Always show play button or `primaryButtonBar` when video paused?
Is there a way to always show `primaryButtonBar` when the video is in pause state? Right now, the controls are hidden where there is no mouse movement for specified `controlsHoverDuration`. As a workaround, I can set the `controlsHoverDuration` to a long time but that still hides the controls when I tap the video. Another way is to build custom video controls but I want to stick to `MaterialVideoControls`. Is there a better way to always show `primaryButtonBar` in paused state using `MaterialVideoControlsTheme`?

fadone:
Resolved #940


alexmercerind:
Why is it saved in State<T>... when is never changed...

PR #994, opened 2024-10-17
fix: catch observe property callback exception

SmartVive:
Property callback exception causes the player to freeze


SmartVive:
Exceptions do not necessarily thrown in 'toDartString'

The callback comes from the 'observeProperty' method. If I call 'observeProperty' and an exception is thrown in the callback, the player will freeze.

alexmercerind:
Fair enough.

alexmercerind:
Still... I want to land that again.

alexmercerind:
Landed that back:

- #998

PR #1003, opened 2024-10-24
Added Player level access to commands `frame-step` and `frame-back-step`

deckerst:


abdelaziz-mahdy:
Thank you for your contribution, is it possible to add tests for these new methods just to make sure they work in all platforms? and keep track if they break for any reason?

deckerst:
I can add treats, but I've added implementation for the "real" player only, not the web player (which seems to be completely different and probably does not support frame stepping). Is that ok?

abdelaziz-mahdy:
I think in that case it will not be added since it will only confuse people 😅 if the support can be added in both we can expose it , other than that we leave it to people to use the set property 

@alexmercerind is what I am saying correct?

deckerst:
Do you mean calling the MPV command "frame-step" directly from the consumer app?

abdelaziz-mahdy:
Yes you can use it from the player instance 

Check the player type then call it 

Example 
```
If (player is nativeplayer)
{
(player as NativePlayer).setproperty();
}
```
Note I am writing the code on mobile so it's 99% wrong syntax wise

deckerst:
In case others need this, here's what I use in the end:
```dart
    final platform = _instance.platform;
    if (platform is NativePlayer) {
      if (frameCount > 0) {
        await platform.command(['frame-step']);
      } else if (frameCount < 0) {
        await platform.command(['frame-back-step']);
      }
    }
```

PR #1007, opened 2024-10-28
build(darwin): add GNU env support

birros:
Hi everyone,

Small updates in the libs installation scripts to support the GNU versions of `sed` and `tar`.

Required for everyone who, like me, chooses to replace macOS native tools with their GNU equivalents.


alexmercerind (approved):
LGTM!

PR #1022, opened 2024-11-05
Fix crashes on systems where DCL is restricted

### LINKED ISSUE #1015: media-kit won't work without DCL
Test app (or any app using media-kit such as Aves, where this was originally [discovered](https://github.com/deckerst/aves/issues/1271)) will not work without DCL, therefore intentionally is crashed by the OS.

cillyvms:
On Linux-based systems using NativeCallables requires creating executable memory mappings (Dynamic Code Loading), otherwise the Dart VM crashes.
This PR adds a check for DCL and fallbacks to the isolate-based event loop, when DCL is disabled.
Plays fine on both my devices, but the isolate event loop had most likely race conditions, with the fix I can't reproduce a crash anymore.

Fixes #1015

### DIFF
--- a/media_kit/lib/src/player/native/core/execmem_restriction.dart
+++ b/media_kit/lib/src/player/native/core/execmem_restriction.dart
@@ -1,6 +1,6 @@
 /// This file is a part of media_kit (https://github.com/media-kit/media-kit).
 ///
-/// Copyright © 2021 & onwards, Hitesh Kumar Saini <saini123hitesh@gmail.com>.
+/// Copyright © 2024 & onwards, cillyvms <cillyvms@estrogen.dev>.
 /// All rights reserved.
 /// Use of this source code is governed by MIT license that can be found in the LICENSE file.
 import 'dart:ffi';
--- a/media_kit/lib/src/player/native/core/execmem_restriction.dart
+++ b/media_kit/lib/src/player/native/core/execmem_restriction.dart
@@ -1,0 +1,45 @@
+/// This file is a part of media_kit (https://github.com/media-kit/media-kit).
+///
+/// Copyright © 2021 & onwards, Hitesh Kumar Saini <saini123hitesh@gmail.com>.
+/// All rights reserved.
+/// Use of this source code is governed by MIT license that can be found in the LICENSE file.
+import 'dart:ffi';
+import 'dart:io';
+
+typedef MmapFunctionNative = Pointer<Void> Function(Pointer<Void> address, Size length, Int32 protection, Int32 flags, Int32 fd, Long offset);
+typedef MmapFunction = Pointer<Void> Function(Pointer<Void> address, int length, int protection, int flags, int fd, int offset);
+
+typedef MunmapFunctionNative = Int32 Function(Pointer<Void> address, Size length);
+typedef MunmapFunction = int Function(Pointer<Void> address, int length);
+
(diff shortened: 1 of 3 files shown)


media_kit/lib/src/player/native/core/execmem_restriction.dart
```diff
@@ -0,0 +1,45 @@
+/// This file is a part of media_kit (https://github.com/media-kit/media-kit).
+///
+/// Copyright © 2021 & onwards, Hitesh Kumar Saini <saini123hitesh@gmail.com>.
```
alexmercerind, line 3: You can have your own name in the license header. This is bit of something that we do:

- https://github.com/media-kit/media-kit/blob/21d2853fd3bfce6a67249850d8ff2aee43d00730/media_kit/lib/src/player/web/utils/hls.dart#L3
- https://github.com/media-kit/media-kit/blob/21d2853fd3bfce6a67249850d8ff2aee43d00730/video_player_media_kit/lib/video_player_media_kit.dart#L3

media_kit/lib/src/player/native/core/initializer_isolate.dart
```diff
@@ -0,0 +1,190 @@
...
+
+    // Sending the address of the created [mpv_handle] & the [SendPort] of the [receivePort].
+    // Raw address is sent as [int] since we cannot transfer objects through Native Ports, only primitives.
+    port.send(handle.address);
+
+    // Lookup for events & send to main thread through [SendPort].
+    // Ensuring the successful sending of the last event before moving to next [MPV.mpv_wait_event].
+    while (!disposed) {
+      completer = Completer();
+      final event = mpv.mpv_wait_event(handle, kReleaseMode ? 1 : 0.1);
```
alexmercerind, line 169: Any value >= 0 essentially means polling. It may slow down Dart's event loop. It may be a good idea to keep -1 & address any issues that it brings some other way.

alexmercerind (commented):
Thanks! I will try to take this forward. The `execmem_restriction.dart` is hot stuff!

media_kit/lib/src/player/native/core/execmem_restriction.dart
cillyvms: Ah, sorry, I just copied the header and didn't think of that. I'll change it.

media_kit/lib/src/player/native/core/initializer_isolate.dart
cillyvms: I agree, this is not a great solution. It seems to have trouble with waking up. I have an idea of what might be going on, I'll see what can be done.

cillyvms:
It looks like the isolate wasn't getting reliably disposed, because it was blocking in mpv's code waiting for an event.
Since the isolate really only runs one thread and incoming messages are handled by the event loop, I made it so the main isolate interrupts the wait instead and the mpv loop isolate yields control to Dart's event loop every iteration, so it has a chance to process the dispose message.

Also is there a reason for the `mpv_terminate_destroy` function being called from the player instead of initializer? This can cause the main isolate to free the mpv handle before the other isolate is done processing events.
https://github.com/media-kit/media-kit/blob/e1c5f5dc8b9c4174c59530bdf192e2f95622f702/media_kit/lib/src/player/native/player/real.dart#L105-L107

alexmercerind (approved):
Thanks a lot for your work!

PR #1036, opened 2024-11-15
perf: optimize playlist state/stream in `NativePlayer`

alexmercerind:
It is too expensive to iterate through returned `mpv_node*`.


alexmercerind:
I broke shuffle in this it seems... will make another pull request.

PR #1046, opened 2024-11-28
fix: Fix playlist loading performances

JulienDev:
When the playlist is big it takes a lot of time to load it. For example, on a Pixel 5, it takes 1200ms to open a playlist with ~6000 medias causing an UI freeze. By simply replacing the current loop with a StringBuilder it only takes around 100ms which is 10 times faster


alexmercerind (approved):
You are legend.

alexmercerind:
We've been able to optimize playlists massively as part of #1036 & this PR.

PR #1062, opened 2024-12-11
Added dynamic headers

### LINKED ISSUE #897: How to set dynamic headers.
I want to send the different headers data each time it asks for new video content, this does work the first time when video loads but when user seek the video which is not buffered it fails as my server need the noune unique each time, this is what i have now.

```
  Map<String, String> headers = {
        HttpHeaders.authorizationHeader: 'Bearer $token',
        ...AppHashingCryptography.getExtraHashedData(),
      };

    
    courseVideoPlayer
        .open(
      Media(
        currentVideo.value!.videoPath,
        httpHeaders: headers,
      ),
      play: true,
    )
        .then((_) {
      // Successfully opened the video
      print('Video opened successfully');
    }).catchError((err) {
      // Handle the error
      print('Error: $err');
    });
```
token always stays the same but the headersData from cryptography needs to be different but it keep sending the first loaded headers which is not acceptable at serverside.

Thanks

LNA-DEV:
I tried to build something to fix #897.

This seems to work in my tests, but I think someone more involved in the codebase should take a look at it.

What I currently think is not optimal is the frequency with which the headers get updated. Maybe another position to call the updateHeaders method would be better?

I think this feature would be really great and look forward to collaborating on it! 😄

### DIFF
--- a/media_kit/lib/src/models/media/media_native.dart
+++ b/media_kit/lib/src/models/media/media_native.dart
@@ -87,6 +87,11 @@ class Media extends Playable {
   /// Default: `null`.
   final Map<String, String>? httpHeaders;
 
+  /// Dynamic HTTP headers.
+  ///
+  /// Default: `null`.
+  final Map<String, String> Function()? dynamicHttpHeaders;
+
   /// Start position.
   ///
   /// Default: `null`.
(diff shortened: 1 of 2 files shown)


LNA-DEV:
I updated the code to only refresh the headers if pause / play is pressed or seek used. Therefore the call frequency is greatly reduced. I think this should be enough for basic usecases because in my experiments the HTTP requests seemed to only close and reopen on pause / seek.

alexmercerind:
Hi! Just make the existing `httpHeaders` "dynamic". Remove existing GC finalizer based logic.

LNA-DEV:
Hej. I am not exactly sure if I understand correctly what your plan is.   
  
You want to update the httpHeaders of the original Media object and the current GC code removes this object? But if it would not be cleaned up you give the pointer to mpv and so the headers would be "dynamic"?

PR #1068, opened 2024-12-24
Fix: improve responsiveness of showing controls on mobile

abdelaziz-mahdy:
this makes the controls come up in instant instead of the sides being slow due to waiting for double tap without breaking double tap


alexmercerind (approved):
LGTM!

Let's land #1070 before.

PR #1074, opened 2025-01-01
fix: replace `EdgeInsets` with `EdgeInsetsGeometry`

kaankoken:
Hello & Happy New Year.

I have just started using your package, and we are giving the RTL support on our app. The `media_kit` package does not behave well with RTL since does not use a general interface, EdgeInsetsGeometry.

This change should not change any behavior. It only changes the interface

### DIFF
--- a/media_kit_video/lib/media_kit_video_controls/src/controls/extensions/padding.dart
+++ b/media_kit_video/lib/media_kit_video_controls/src/controls/extensions/padding.dart
@@ -1,0 +1,40 @@
+import 'package:flutter/widgets.dart';
+
+extension EdgeInsetsGeometryExt on EdgeInsetsGeometry? {
+  double get bottom {
+    final value = this;
+
+    if (value == null) return 0.0;
+    if (value is EdgeInsets) return value.bottom;
+    if (value is EdgeInsetsDirectional) return value.bottom;
+
+    return 0.0;
+  }
+
+  double get top {
+    final value = this;
+    if (value == null) return 0.0;
+    if (value is EdgeInsets) return value.top;
+    if (value is EdgeInsetsDirectional) return value.top;
+
+    return 0.0;
+  }
+
+  double get start {
+    final value = this;
+    if (value == null) return 0.0;
+    if (value is EdgeInsets) return value.left;
+    if (value is EdgeInsetsDirectional) return value.start;
+
+    return 0.0;
+  }
+
+  double get end {
+    final value = this;
+    if (value == null) return 0.0;
+    if (value is EdgeInsets) return value.right;
+    if (value is EdgeInsetsDirectional) return value.end;
+
+    return 0.0;
+  }
+}
--- a/media_kit_video/lib/media_kit_video_controls/src/controls/material.dart
+++ b/media_kit_video/lib/media_kit_video_controls/src/controls/material.dart
@@ -192,7 +192,7 @@ class MaterialVideoControlsThemeData {
   /// * FullScreen: `MediaQuery.of(context).padding`
   ///
   /// NOTE: In fullscreen, this will be safe area (set [padding] to [EdgeInsets.zero] to disable safe area)
-  final EdgeInsets? padding;
+  final EdgeInsetsGeometry? padding;
 
   /// [Duration] after which the controls will be hidden when there is no mouse movement.
   final Duration controlsHoverDuration;
@@ -224,13 +224,13 @@ class MaterialVideoControlsThemeData {
   final List<Widget> topButtonBar;
 
   /// Margin around the top button bar.
-  final EdgeInsets topButtonBarMargin;
+  final EdgeInsetsGeometry topButtonBarMargin;
 
   /// Buttons to be displayed in the bottom button bar.
   final List<Widget> bottomButtonBar;
 
   /// Margin around the button bar.
-  final EdgeInsets bottomButtonBarMargin;
+  final EdgeInsetsGeometry bottomButtonBarMargin;
 
   /// Height of the button bar.
   final double buttonBarHeight;
@@ -244,7 +244,7 @@ class MaterialVideoControlsThemeData {
   // SEEK BAR
 
   /// Margin around the seek bar.
-  final EdgeInsets seekBarMargin;
+  final EdgeInsetsGeometry seekBarMargin;
 
   /// Height of the seek bar.
   final double seekBarHeight;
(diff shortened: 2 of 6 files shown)


abdelaziz-mahdy:
reading this docs https://api.flutter.dev/flutter/painting/EdgeInsetsGeometry-class.html 

so i am confused why this change is needed?
image

kaankoken:
I am not able to use `EdgeInsetsDirectional` since `EdgeInsets` is assigned. Also, it is not the same

```md
An immutable set of offsets in each of the four cardinal directions, but whose horizontal components are dependent on the writing direction.

This can be used to indicate padding from the left in [TextDirection.ltr](https://api.flutter.dev/flutter/painting/dart-ui/TextDirection.html) text and padding from the right in [TextDirection.rtl](https://api.flutter.dev/flutter/painting/dart-ui/TextDirection.html) text without having to be aware of the current text direction.
```
https://api.flutter.dev/flutter/painting/EdgeInsetsDirectional-class.html

kaankoken:
when you use `EdgeInsets` your ui only supports one direction which is `LTR`.  when you use `EdgeInsetsDirectional`, you do not have to take care the text direction @abdelaziz-mahdy

abdelaziz-mahdy:
oh, all good will leave it to alex to make sure all is good,

kaankoken:
@abdelaziz-mahdy maybe you should consider to change(or I can open an separate PR if It is okay) for `Alignments` as well

abdelaziz-mahdy:
my main problem is that change may break other people code, and as you can see from the workflows, the example fails

abdelaziz-mahdy:
actually looking at the error, its coming from the package itself, so it needs to be fixed there first, please make sure there is no errors then we can review it

kaankoken:
> my main problem is that change may break other people code, and as you can see from the workflows, the example fails

`EdgeInsetsGeometry` is indeed a general `interface`. I did not get it why its failed. I will check it out

alexmercerind (approved):
LGTM

abdelaziz-mahdy:
> LGTM

 its not working correctly, there is errors on the builds, it should not be merged until we resolve these, also it will be a breaking change since if someone is depending on the themes edge insect they will need to resolve it first before doing whatever they were doing

kaankoken:
> s not working correctly, there is errors on the builds, it should not be merged until we resolve these, also it will be a breaking change since if someone is depending on the the

I think I solved the issues. I had to create an extension to solve it.

abdelaziz-mahdy:
i am fine with the extension, but still it will break other people code, if they are depending on the theme directly, so i am thinking of doing getters that will return the resolved EdgeInsets, that will keep the interface the same, and give everyone the ability to pass the object they prefer.

what do you think?

kaankoken:
> i am fine with the extension, but still it will break other people code, if they are depending on the theme directly, so i am thinking of doing getters that will return the resolved EdgeInsets, that will keep the interface the same, and give everyone the ability to pass the object they prefer.
> 
> what do you think?

This should not break anybody's code. When you introduce `EdgeInsetsGeometry`,  you could either use `EdgeInsets` or `EdgeInsetsDirectional`. It will keep working the same as it was before. Maybe, the people did not care much about the `RTL` version of the app or they did not support it at all.

abdelaziz-mahdy:
> > i am fine with the extension, but still it will break other people code, if they are depending on the theme directly, so i am thinking of doing getters that will return the resolved EdgeInsets, that will keep the interface the same, and give everyone the ability to pass the object they prefer.
> > what do you think?
> 
> This should not break anybody's code. When you introduce `EdgeInsetsGeometry`, you could either use `EdgeInsets` or `EdgeInsetsDirectional`. It will keep working the same as it was before. Maybe, the people did not care much about the `RTL` version of the app or they did not support it at all.

if someone was using the padding object from our class in their app, it will now return `EdgeInsetsGeometry` instead of `EdgeInsets` same as inside of our package, they may need to do stuff like you did, thats a breaking change from what i know


since the public interface has changed its safer to assume that it may break people codes,


and i am saying i dont mind the change, but if we can maintain the same interface that will be awesome

abdelaziz-mahdy:
@moffatman what do you think, does changing the interface better or maintaining it behind a getter of some kind (note, we may need to pass the direction of text to the getter)

i think changing the interface is better, but since its a breaking change it will need to be mentioned how to fix it i think

kaankoken:
@abdelaziz-mahdy  I have created an example. Could you look into it?  could you tell me if I missing something?

[example](https://github.com/kaankoken/media-kit/blob/example/edge_insets_geometry/media_kit_video/example/main.dart)
You could look into line 39 which uses `topButtonBarMargin` & line 58 which uses `bottomButtonBarMargin`. this does not break anything, or am I missing something?

abdelaziz-mahdy:
> @abdelaziz-mahdy I have created an example. Could you look into it? could you tell me if I missing something?
> 
> [example](https://github.com/kaankoken/media-kit/blob/example/edge_insets_geometry/media_kit_video/example/main.dart) You could look into line 39 which uses `topButtonBarMargin` & line 58 which uses `bottomButtonBarMargin`. this does not break anything, or am I missing something?

It does break the interface, but not from the constructor’s side—rather, from the retrieval side.

This is why you had to create the extension. Normally, the paddings were returned as `EdgeInsets`, but returning them as `EdgeInsetsGeometry` removes all the extra functions (like the `+`, `-`, `/`, `*` operators, among others).

If it requires more changes to the package internals (beyond just changing the types), then it would be a breaking change if the interface is public. My concern is that I’m not sure if anyone is actually using that padding from our package. If no one is using it (or if it’s not meant to be used externally), we can safely merge it. 

this is why i am waiting for a second review to confirm if my concerns are true, or we should ignore them.

abdelaziz-mahdy:
example someone may be doing this for any reason
``` 
           key.currentState?.videoViewParametersNotifier.value
                          .subtitleViewConfiguration.padding+EdgeInsets.all(10);
```

which will not work after we merge this and will require some changes

is it a common use case, i dont think so, but we need to keep it in mind so we mention it for people updating

moffatman:
Can we have an automatic getter? With direction from context?

kaankoken:
> Can we have an automatic getter? With direction from context?

I couldn't picture it in my head. Could you provide an example so that I could add into this PR

moffatman:
Wait. why can't you just resolve the direction when you construct the controls?

`MaterialVideoControlsThemeData(padding: EdgeInsetsDirectional.only(start: 56.0).resolve(Directionality.of(context)))`

And we don't need this change?

---

The getter idea would be to store the `padding` as a private variable. Then getter would be like


But we could always just do a breaking change.

PR #1077, opened 2025-01-06
chore: refactor mimalloc import

TheOneWithTheBraid:
In some cases, users of the plugin might prefer to use a dynamically linked version of mimalloc. This PR introduces a new CMake option allowing to dynamically link mimalloc. Unlike #883 , this PR does **not** change the default behavior but simply adds an additional CMake option.

**The default behavior of the plugin is not changed.**

If users set the CMake option :

```cmake
set(MIMALLOC_USE_STATIC_LIBS OFF)
```

This plugin will now link against the system-provided mimalloc.

Users might prefer this in case they try to reduce installation size or aim to comply with distributor-specific dependency management requirements.

Unless this option is set to `OFF`, the behavior of the plugin won't change.

Thanks a lot for your amazing work 🎉 !

### DIFF
--- a/README.md
+++ b/README.md
@@ -1683,6 +1683,21 @@ This is as simple as [adding one line to `linux/CMakeLists.txt`](https://github.
 target_link_libraries(${BINARY_NAME} PRIVATE ${MIMALLOC_LIB})
 ```
 
+In case you prefer dynamic linking of mimalloc, you can additionally add the following line to your `linux/CMakeLists.txt`:
+
+```cmake
+# use dynamically linked mimalloc
+set(MIMALLOC_USE_STATIC_LIBS OFF)
+```
+
+In this case, please ensure you install `libmimalloc-dev` at compile time and `libmimalloc2.0` as runtime dependencies.
+
+#### Ubuntu/Debian
+
+```bash
+sudo apt install libmimalloc-dev libmimalloc2.0
+```
+
 ### Web
 
 On the web, **libmpv is not used**. Video & audio playback is handled by embedding [HTML `<video>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video). The format support depends upon the web browser. It happens to be extremely limited as compared to native platforms.
--- a/libs/linux/media_kit_libs_linux/CHANGELOG.md
+++ b/libs/linux/media_kit_libs_linux/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.2.0
+
+- chore: refactor mimalloc import
+
 ## 1.1.3
 
 - feat: improve compile-time failure handling
--- a/libs/linux/media_kit_libs_linux/linux/CMakeLists.txt
+++ b/libs/linux/media_kit_libs_linux/linux/CMakeLists.txt
@@ -4,15 +4,16 @@
 # All rights reserved.
 # Use of this source code is governed by MIT license that can be found in the LICENSE file.
 
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.13)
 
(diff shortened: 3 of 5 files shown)


alexmercerind:
I am a bit curious... Why do you want to switch to the mimalloc package from the distribution?

Eventually, we want to stop depending even on mpv package from distribution. Are you aware a number of issues can come simply by changing versions e.g. you are testing your app against version A of mimalloc & your users run your app against version B or C of mimalloc. On the other hand, there can be breaking changes too.

Simply, do not do this.

TheOneWithTheBraid:
Well, I guess these are two ground-up different philosophies : Whereas I can see a huge advantage in static linking for individual projects with its own distribution system (e.g. binaries on GitHub Releases or website), as developer of projects aiming to integrate deeper into the Linux desktop ecosystem, I highly prefer to have the distribution-provided packages with externally maintained dependencies.

In my case, I'm targeting both GLibC distributions as well as libmusl distributions. Statically linking a particular version is impossible here. I rely on the distributions to package and automatically test against library versions. I can therefore comply with distribution best-practices such as particular library features being turned off or on - as well as on security updates provided from distribution side.

Sure, many people might not develop Flutter applications as "traditional" Linux packages, but I definitely see a huge advantage of at least providing the opportunity to develop Flutter apps using media_kit as typical Linux packages.

In case you migrate to a vendored / statically linked version of MPV, I'd highly appreciate to provide a CMake option like `MPV_USE_STATIC_LIBS` in order to stick at the distribution-provided MPV version at own risk - we will otherwise no longer be able to ship applications making use of media_kit as native packages on Flutter-supported desktop distributions (which is Alpine and Arch Linux as of now).

TheOneWithTheBraid:
As of now, on [Alpine Flutter packages using media_kit](https://gitlab.alpinelinux.org/alpine/aports/-/tree/master/testing/polycule), we already *do* require to depend on `media_kit_libs_linux` from the patched version from this PR and additionally need to ship [musl-related patches for `media_kit_video`](https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/polycule/media-kit-broken-by-design.patch.pubdev) - in future, I'd appreciate to be able to ship media_kit as it without patches on distribution and application side, maybe you can understand this effort.

Therefore the proposa : let's keep it statically linked by default but at least provide an option for maintainers to simply use a dynamically linked version instead.

alexmercerind:
I think your situation is very specific, and you would be better off maintaining your patches in a private repository. For instance, [this](https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/polycule/media-kit-broken-by-design.patch.pubdev#L7-10) does not look normal to me.

Tagging @moffatman for second opinion.

TheOneWithTheBraid:
> I think your situation is very specific, and you would be better off maintaining your patches in a private repository. For instance, [this](https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/polycule/media-kit-broken-by-design.patch.pubdev#L7-10) does not look normal to me.

As mentioned, this is a musl-specific patch - which we therefore do not aim to upstream. The `NULL` definition of media_kit was different from the libmusl expected one. But that's not relevant for this PR.

Since dynamic linking is generally an advantage on GLibC distributions too, I would not consider this as "very specific". Major players such as Canonical (the makers of Ubuntu) count on Flutter for desktops. Whatever advantages one might personally see with static linking, the way to go on Linux desktops is dynamic linking in a traditional Linux ecosystem. Again, I do not ask to dynamically link by default but rather to at least provide maintainers the opportunity to do so.

alexmercerind (approved):
LGTM!

PR #1107, opened 2025-02-15
fix(android): use video-params instead of video-output-params

### LINKED ISSUE #1091: [Android] Video output broken with mpv 0.37 or higher
Currently, on the Android platform, we use libmpv.so binary from https://github.com/media-kit/libmpv-android-audio-build. We are using version 1.1.5 (based on mpv 0.36). 

However, if we switch to the 1.1.6 release from the same repository, the video starts flickering constantly. I tried manually compiling other versions of libmpv, and any version later than mpv 0.3.7 triggers the same issue. Since mpv 0.36 is outdated, it might be better to fix this issue or at least remove the broken 1.1.6 release from https://github.com/media-kit/libmpv-android-audio-build.

Predidit:
Closes #1091

### DIFF
--- a/media_kit/lib/src/player/native/player/real.dart
+++ b/media_kit/lib/src/player/native/player/real.dart
@@ -1928,7 +1928,7 @@ class NativePlayer extends PlatformPlayer {
           }
         }
       }
-      if (prop.ref.name.cast<Utf8>().toDartString() == 'video-out-params' &&
+      if (prop.ref.name.cast<Utf8>().toDartString() == 'video-params' &&
           prop.ref.format == generated.mpv_format.MPV_FORMAT_NODE) {
         final node = prop.ref.data.cast<generated.mpv_node>().ref;
         final data = <String, dynamic>{};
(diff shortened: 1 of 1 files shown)


Predidit:
Seems that Linux CI already broken on branch master.

alexmercerind:
I'm not fully convinced how this works. Do `video-out-params` & `video-params` return different value? Is it a race condition?

Predidit:
@alexmercerind 

Before mpv 0.37, both `video-out-params` and `video-params` returned identical values. However, after mpv 0.37, the behavior has changed.

When testing with the first video example in media_kit_test, the `width` value in `video-out-params` continuously fluctuates between 960 and 854. Each fluctuation triggers a resize event, causing persistent flickering.

Meanwhile, `video-params` correctly reports the width as 960, consistent with the behavior in mpv versions prior to 0.37.

This inconsistency between the two parameters and the constant width fluctuation in video-out-params is causing display problems that didn't exist in earlier versions.

abdelaziz-mahdy:
from their docs 
image
do you use any video filters? if not they should be the same :), do we use any video filters? i didnt see that we specify any in our code (maybe i missed it?)

Predidit:
@abdelaziz-mahdy 

I have also checked the documentation for this part, but the flicker issue casued by unexpected resize event does exist. You can reproduce the issue with libmpv-android-build 1.1.6, the fix in this PR is effective.

abdelaziz-mahdy:
> @abdelaziz-mahdy
> 
> I have also checked the documentation for this part, but the flicker issue casued by unexpected resize event does exist. You can reproduce the issue with libmpv-android-build 1.1.6, the fix in this PR is effective.

mmmmm, since i dont not understand mpv that much i cant be sure why it was implemented this way from the start and dont know if changing it will break the video for other people, so i will wait for alex

Issue #1127, opened 2025-03-02
Media kit not compatible with some other plugins due to js version


Jonny1987:
It it possible to update the js package since it conflicts with come other plugins:

For example:
```
Because stream_chat_flutter 9.4.0 depends on media_kit ^1.1.10+1 which depends on js ^0.6.7, stream_chat_flutter 9.4.0 requires js ^0.6.7.
And because amplitude_flutter >=3.16.5 <4.0.0-beta.1 depends on js ^0.7.0, stream_chat_flutter 9.4.0 is incompatible with amplitude_flutter >=3.16.5 <4.0.0-beta.1.
So, because workwith depends on both amplitude_flutter ^3.16.7 and stream_chat_flutter 9.4.0, version solving failed.
```

Apparently the js package is also discontinued now anyway and js_interop should be used: https://pub.dev/packages/js

EricApostal:
I actually believe being a dependency at all is a mistake. We have already migrated to js_interop / web for wasm support. I'll give this a go later today.

EricApostal:
Should be fixed at https://github.com/media-kit/media-kit/pull/1128

PR #1128, opened 2025-03-04
Remove unused js dependency

EricApostal:
# Issue
Keeping JS imported conflicts with other packages. This is not longer used.

Fixes https://github.com/media-kit/media-kit/issues/1127


abdelaziz-mahdy (approved):
LGTM

PR #1144, opened 2025-03-24
regain focus when media begins to play fixes #496

### LINKED ISSUE #496: focus on video
Hi
I want to focus on Video by Dpad. Indeed I want to show Material Buttons by click on Dpad.
Is There any solution to it?

james-lawrence:
ideally the focus node would be provided by the Player and would be properly managed during methods but this resolves the issue for focus being lost and never regained when the focus is removed by the application.


james-lawrence:
closing for a bit; need to readjust and finish testing. there are issues in the main branch vs the latest publish version.

PR #1145, opened 2025-03-24
regain focus when media begins to play fixes #790

### LINKED ISSUE #790: Keyboard shortcuts disabled once the focus is moved to other widgets
## Description

Unable to move focus back to Video once it's moved to other widgets, thus making keyboard shortcuts disabled.

## Reproduce code

```
import 'package:flutter/material.dart';
import 'package:media_kit/media_kit.dart'; // Provides [Player], [Media], [Playlist] etc.
import 'package:media_kit_video/media_kit_video.dart'; // Provides [VideoController] & [Video] etc.

void main() {
  WidgetsFlutterBinding.ensureInitialized();
  // Necessary initialization for package:media_kit.
  MediaKit.ensureInitialized();
  runApp(
    const MaterialApp(
      home: MyScreen(),
    ),
  );
}

class MyScreen extends StatefulWidget {
  const MyScreen({Key? key}) : super(key: key);
  @override
  State<MyScreen> createState() => MyScreenState();
}

class MyScreenState extends State<MyScreen> {
  // Create a [Player] to control playback.
  late final player = Player();
  // Create a [VideoController] to handle video output from [Player].
  late final controller = VideoController(player);

  @override
  void initState() {
    super.initState();
    player.open(Media(
        'https://user-images.githubusercontent.com/28951144/229373695-22f88f13-d18f-4288-9bf1-c3e078d83722.mp4'));
  }

  @override
  void dispose() {
    player.dispose();
    super.dispose();
  }

  @override
  Widget build(BuildContext context) {
    return Row(
      children: [
        Expanded(
            flex: 2,
            child: Video(
              controller: controller,
            )),
        Expanded(
            child: Container(
          child: TextField(),
        )),
      ],
    );
  }
}
```

Reproduce steps:

1. Test keyboard shortcuts upon build such as space to pause/play the video, it's normal at this step.
2. Click the TextField to move focus to it.
3. Click the Video widget again and try to use keyboard shortcuts, now they wouldn't work unless fullscreen mode is entered.

james-lawrence:
resolves not regaining focus while media is playing. fixes #790

### DIFF
--- a/media_kit_video/lib/media_kit_video_controls/src/controls/material_desktop.dart
+++ b/media_kit_video/lib/media_kit_video_controls/src/controls/material_desktop.dart
@@ -372,6 +372,7 @@ class _MaterialDesktopVideoControlsState
   DateTime last = DateTime.now();
 
   final List<StreamSubscription> subscriptions = [];
+  final FocusNode _focus = FocusNode();
 
   double get subtitleVerticalShiftOffset =>
       (_theme(context).padding?.bottom ?? 0.0) +
@@ -407,6 +408,11 @@ class _MaterialDesktopVideoControlsState
               });
             },
           ),
+        controller(context).player.stream.playing.listen((playing) {
+          // regain focus when media begins playing.
+          if (!playing) return;
+          _focus.requestFocus();
+        }),
         ],
       );
 
@@ -558,6 +564,7 @@ class _MaterialDesktopVideoControlsState
                   exitFullscreen(context),
             },
         child: Focus(
+          focusNode: _focus,
           autofocus: true,
           child: Material(
             elevation: 0.0,


alexmercerind:
It's hard to approve this one.

It is not a good idea to modify the existing focus each time a new media starts playing. This is not a behavior that everyone will want. In case someone does, they can add their own `Focus` widget (on top of `Video`) & control it using that. But, by introducing this behavior, everyone is forced into this choice (which may introduce unwanted behavior for them).

james-lawrence:
*shrug* it works and is probably closer to the expected behavior than the current state. but this is a pretty irritating issue with no decent workaround afaik. if the focus node was injectable it wouldn't be a issue. maybe expose a way to inject it so people can actually manage the focus as they wish in their app?

abdelaziz-mahdy:
what about having the same pr, but without the stream part related to focus, but we expose the internal focus node so people can have their own behaviour?

james-lawrence:
would be happy to do it if i was in a place where i had the time; currently have other priorities. opened just to show a path forward to be honest.

Issue #1167, opened 2025-04-18
Playlist Updates Incorrectly in Certain Scenarios


RiverAge:
### Description

In certain scenarios, when a media item is removed from the Playlist, the onData callback output is not as expected. The onData callback should reflect the updated length of the media list, but the current behavior is incorrect.

###  Steps to Reproduce

```dart
final player = Player();
final playlist = Playlist([ Media('https://file_example_MP3_700KB.mp3'), Media('https://file_example_MP3_1MG.mp3'), ], index: 1);
await player.open(playlist, play: false);
await player.setPlaylistMode(PlaylistMode.loop);
player.stream.playlist.listen((onData) {
  log('onData called ${onData.medias.length}');
});
await player.remove(1);
```

###  Expected Behavior
The onData callback should output:
```
[log] onData called 1
```

### Actual Behavior
The onData callback outputs:

```
[log] onData called 1
[log] onData called 2
````

### Possible Cause

 after digging the code, I suspect the problem lies in the following section of the code:
https://github.com/media-kit/media-kit/blob/e8ef211092c0259012d52213031738416ce6ab3d/media_kit/lib/src/player/native/player/real.dart#L496

It appears that the current variable is updated in the default handling case. However, in certain situations, such as when removing items from the playlist, current is not updated. This discrepancy seems to cause the _handler callback to use an outdated value for current, leading to the incorrect update reflected in the onData callback. no idea why `_handler` gets called after `playlistController.add` in the remove function

https://github.com/media-kit/media-kit/blob/e8ef211092c0259012d52213031738416ce6ab3d/media_kit/lib/src/player/native/player/real.dart#L1595

alexmercerind:
We invite your changes. We do have some unit-tests, which hopefully test regressions... they are not "complete" however.

RiverAge:
Thank you for your response. I will try to submit a Pull Request to address this issue.

PR #1168, opened 2025-04-18
fix:Playlist Updates Incorrectly in Certain Scenarios

RiverAge:
`current` value not synchronously  updated with `playlist` state which leads to potential issues.

fix: #1167


alexmercerind:
Thanks, most `player-remove-*` tests still pass. May you make these similar changes to the `WebPlayer` impl. too?

RiverAge:
> Thanks, most `player-remove-*` tests still pass. May you make these similar changes to the `WebPlayer` impl. too?

Sure, glad to do it.

alexmercerind (approved):
LGTM!

PR #1181, opened 2025-05-06
fixes for URI normalization

Zekfad:
this removes unnecessary Uri.parse calls (which throws with obscure URIs such as `lavf://crypto:file:D:/file.mp3.enc`)
and changes schemes check to ignore scheme casing


Zekfad:
@alexmercerind may I know why this is closed? This is the feature we use in our project, is there something missing I should address for this to be accepted?

PR #1197, opened 2025-05-26
Fixed the issue that playback did not actually start when playing is …

copywith9:
1. Play more than 2 audios first.
2. Then pause the playback.
3. Then drag the progress bar to let the Player automatically go to the next.

Then the Player is not playing, but `playing` == true

This PR fixes it.


oldHarvester:
Please check this PR as faster as possible. This is horrible bug

fxpayback:
please check the PR as I got same issue @alexmercerind

PR #1247, opened 2025-08-22
fix: Ensure pixel_buffer is 4-bits aligned

wenyue:
This PR fix two bugs:
1) Make sure pixel_buffer is 4-bits aligned.
2) Fix texture size caculation wrong.

### DIFF
--- a/media_kit_video/linux/include/media_kit_video/texture_sw.h
+++ b/media_kit_video/linux/include/media_kit_video/texture_sw.h
@@ -37,6 +37,6 @@ gboolean texture_sw_copy_pixels(FlPixelBufferTexture* texture,
 #define SW_RENDERING_MAX_WIDTH 1920
 #define SW_RENDERING_MAX_HEIGHT 1080
 #define SW_RENDERING_PIXEL_BUFFER_SIZE \
-  (SW_RENDERING_MAX_WIDTH) * (SW_RENDERING_MAX_HEIGHT) * (4)
+  (SW_RENDERING_MAX_WIDTH) * (SW_RENDERING_MAX_HEIGHT)
 
 #endif  // TEXTURE_SW_H_
--- a/media_kit_video/linux/video_output.cc
+++ b/media_kit_video/linux/video_output.cc
@@ -19,7 +19,7 @@ struct _VideoOutput {
   GObject parent_instance;
   TextureGL* texture_gl;
   GdkGLContext* gdk_gl_context;
-  guint8* pixel_buffer;
+  guint32* pixel_buffer;
   TextureSW* texture_sw;
   GMutex mutex; /* Only used in S/W rendering. */
   mpv_handle* handle;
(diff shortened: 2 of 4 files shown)


andrew-tpz:
Texture size caculation in s\w rendering is wrong for portrait videos.

The reason is in these defines:
```c++
// media_kit_video/linux/include/media_kit_video/texture_sw.h
// media_kit_video/windows/video_output.cc
#define SW_RENDERING_MAX_WIDTH 1920
#define SW_RENDERING_MAX_HEIGHT 1080

```

Logs output for HW (vertical video):
```
media_kit: VideoOutput: Using H/W rendering.
...
{handle: 3243713639504, id: 3243673479824, rect: {left: 0, top: 0, width: 1080, height: 1440}}  <-- full height
```
Logs output for SW (the same video): 
```
media_kit: VideoOutput: Using S/W rendering.
...
{handle: 3243713668688, id: 3243617065488, rect: {left: 0, top: 0, width: 1080, height: 1080}}  <-- cropped by height

```

wenyue:
Why is this happening? In this case 1080(video width) * 1080(SW_RENDERING_MAX_HEIGHT) / 1440(video height), the width should be 810, not 1080.

andrew-tpz:
My primary problem is the player regularly falls back to software rendering — when playing the same video multiple times. Wrong size calculation is just indicator of s/w fallback for me — I haven't investigated it.

I’m using a monitor in portrait mode (1080x1920) on Ubuntu 22.04. The video is scaled to fit the screen width (1080), so its height should be 1440, but it limits to 1080 . I was hoping this PR would help, but unfortunately it didn’t.

As a test, I manually increased SW_RENDERING_MAX_HEIGHT to 1920 — and that helps (with height, not with fallback).

alexmercerind:
That is intentional & correct. The width and height get clamped to 1080p (while maintaining aspect ratio) in software rendering on all platforms. It's not easy job for the CPU to render into that large pixel buffer without GPU acceleration.

alexmercerind:
Your video is likely made portrait via rotation metadata, which if I recall correctly is kinda buggy in software rendering due to a bug in mpv.

alexmercerind:
What exactly is this pull request solving? Why does the stride of the allocated buffer matter? You could save it as void*. We are not accessing it anywhere within our code, it's almost an opaque type, only consumed within mpv & Flutter texture APIs.

andrew-tpz:
> That is intentional & correct. The width and height get clamped to 1080p (while maintaining aspect ratio) in software rendering on all platforms. It's not easy job for the CPU to render into that large pixel buffer without GPU acceleration.

I agree regarding the performance cost — software rendering puts about twice the load on my CPU compared to hardware rendering.

However, there are Shorts, Reels, etc. — meaning the video resolution can be 1080x1920. No extra memory needed if you can swap SW_RENDERING_MAX_HEIGHT and SW_RENDERING_MAX_WIDTH for portrait videos.

P.S. I'm trying to figure out why media_kit stopped using hardware rendering in Flutter 3.32, while it still works when I switch back to Flutter 3.24. But this is probably need a separate issue.

wenyue:
> What exactly is this pull request solving? Why does the stride of the allocated buffer matter? You could save it as void*. We are not accessing it anywhere within our code, it's almost an opaque type, only consumed within mpv & Flutter texture APIs.

According to the MPV document, pixel_buffer must be 4-bits aligned when using the `rgb0` format. Also, the previous texture size calculation was incorrect in software decoding.

PR #1283, opened 2025-10-04
refactor(android): AndroidVideoController impl

alexmercerind:
- Remove unnecessary complexity.
- Fix an issue that caused video output to be stuck on black screen randomly.
- Ensure libmpv does not fallback to `mediacodec-copy` (if `vo=auto` is used) during various operations e.g. opening/switching media in `Player`.

```
PlayerLog(prefix: vd, level: info, text: Using hardware decoding (mediacodec).)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ migrate_working_dir/
 *.ipr
 *.iws
 .idea/
+.cxx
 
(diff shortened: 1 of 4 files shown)


Predidit:
LGTM

test on andorid15 device, and works very well.

Predidit (approved):

Issue #1278, opened 2025-09-26
A Windows application made with Flutter that uses media-kit has abnormally high CPU and memory usage


modai07:
Previously, I developed a Windows application using Tauri 2, which utilized the native WebView2 for video playback. When playing the same video, its CPU usage ranged from 0% to 1% and memory usage was 200MB.
When opening this video with PotPlayer, the CPU usage was also 0% and the memory usage was 180MB.
However, when integrating media-kit with Flutter to play the same video, the CPU usage reaches 18% and the memory usage is 400MB.
I haven’t manually configured to disable hardware acceleration (and I don’t think this issue is related to hardware acceleration).
I wonder if other developers have encountered the same situation? Is this the case on your end too?

srix55:
I have... there's some nasty bug somewhere.. that's leaking memory gradually. Not able to find it though... i have spent weeks on this. I suppose one should get a grok4 or claude pro subscription and feed the entire codebases of media kit & ffmpeg to sort this out. I'm all out of ideas.. had tried every combination of settings.

alexmercerind:
https://github.com/media-kit/media-kit/issues/296#issue-1816920466

Predidit:
@alexmercerind 

Alex, I think this might be related to how we're currently using libmpv render. We can currently only use non-zero-copy decoding methods, which likely introduces some performance overhead. I noticed in the mpv community that they can achieve this on Qt, but the specific details are somewhat quirky and lack portability, as it requires setting some properties for the Qt engine in addition to libmpv code. Do you think this approach is correct? If so, I'll continue moving forward in this direction to see if there are any possible solutions.

Predidit:
@srix55 

If you haven't actually encountered situations where the program was terminated due to running out of memory, then there generally isn't a memory leak. If you notice memory usage continuously increasing while watching videos (even when a shorter cache duration is set), this is usually normal behavior and sometimes hardware-related. This is more like memory being reserved and will be automatically released when memory becomes scarce - we haven't lost control over it.

alexmercerind:
@Predidit 

> I think this might be related to how we're currently using libmpv render. We can currently only use non-zero-copy decoding methods, which likely introduces some performance overhead.

Yes, you are correct. media_kit on Windows also falls back to using `-copy` `hwdec`, primarily because GL implementation provided by ANGLE does not support all features. So, mpv cannot use the zero-copy `hwdec`.

[...]

PR #1286, opened 2025-10-07
refactor(windows): dxd11 texture no copy

Predidit:
part of #1278

This PR simplifies the Windows implementation by completely removing the intermediate `internal_d3d_11_texture_2D_` texture, thereby avoiding GPU-to-GPU copies. libmpv now renders via ANGLE directly into the texture bound to the Flutter Texture. While this still isn’t truly zero-copy—we still can’t use mpv’s hardware decoders without the copy flag—it’s a significant step forward and delivers tangible performance improvements.

### DIFF
--- a/media_kit_video/windows/angle_surface_manager.cc
+++ b/media_kit_video/windows/angle_surface_manager.cc
@@ -59,8 +59,6 @@ void ANGLESurfaceManager::Draw(std::function<void()> callback) {
 void ANGLESurfaceManager::Read() {
   ::WaitForSingleObject(mutex_, INFINITE);
   if (d3d_11_device_context_ != nullptr) {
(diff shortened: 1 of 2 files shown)


alexmercerind:
I don't think this makes any performance differences. This is a GPU to GPU copy, which is being done to prevent black flickering on many Windows devices especially ones with high refresh rates.

This technique is called "Double Buffering" and used very often in Graphics and/or OpenGL.

alexmercerind:
The `-copy` in mpv's hwdec refers to a GPU to CPU or CPU to GPU copy (expensive)... i.e. requiring some sort of CPU emulation for a particular instruction. At least, this is my understanding.

Predidit:
I don’t have enough background in computer graphics. Thanks for pointing out my mistake. I’ll keep investigating this and try to make libmpv use the rendering context we provide directly.

PR #1292, opened 2025-10-12
fix(linux): fix linux hardware render flick

Predidit:
This PR has two main parts:

1. `texture_gl.cc`: Saving the OpenGL state before rendering and restoring it afterward, which resolves the flickering issue.
2. `video_output.cc`: Addresses the frequent freeze on player exit introduced after the `texture_gl.cc` changes by ensuring mpv-related texture resources are disposed in the correct OpenGL context.

### DIFF
--- a/media_kit_video/linux/texture_gl.cc
+++ b/media_kit_video/linux/texture_gl.cc
@@ -31,10 +31,9 @@ static void texture_gl_init(TextureGL* self) {
 
 static void texture_gl_dispose(GObject* object) {
   TextureGL* self = TEXTURE_GL(object);
-
   // Only delete OpenGL resources if there is a valid OpenGL context.
   gboolean deleted = FALSE;
-  if(self->video_output != NULL){
+  if (self->video_output != NULL) {
     GdkGLContext* ctx = video_output_get_gdk_gl_context(self->video_output);
     if (ctx != NULL) {
       gdk_gl_context_make_current(ctx);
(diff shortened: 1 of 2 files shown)


Predidit:
@alexmercerind 

Alex, I’m now fairly certain we can’t use zero-copy hardware decoding on Windows, so I’m trying to enable it on Linux instead. 

From your code, I see you pass the X11/Wayland display to mpv’s renderer, which is required for zero-copy VA-API. Your comments also indicate this was done to make zero-copy VA-API work. VAAPI-copy doesn’t need these parameters, but at the moment we can only use vaapi-copy instead of vaapi. Could you confirm whether zero-copy VA-API (vaapi) was working correctly when you wrote this code?

alexmercerind:
Its actually not possible due to GTK's problems: https://github.com/celluloid-player/celluloid/issues/781#issuecomment-1674345669

I am not really concerned with zero-copy at this point, since auto-safe (which is default in mpv & media_kit) anyway recommends copy hwdec to be compatible on all devices. I am not really finding the difference worth pursuing.

> I’m now fairly certain we can’t use zero-copy hardware decoding on Windows.

Regardless, I'm more hopeful about Windows & macOS than GNU/Linux. GTK is dead end.

alexmercerind:
It appears that someone has actually built DXGI API for Windows:

https://github.com/mpv-player/mpv/compare/master...dragonflylee:mpv-player:d3d11/0.40

alexmercerind:
I will try to experiment next weekend with this.

PR #1294, opened 2025-10-13
fix(android): seek to current playback position instead of start

ashiwalia:
Modified AndroidVideoController.widListener to seek to the current playback position instead of Duration.zero. This improves the UX by avoiding an unexpected rewind.


Joseph-Nathan:
[android] after update 
when i set media to player with start (duration)
the media start from beginning ,
Media.start (Duration) not work ! 

[related issue](https://github.com/media-kit/media-kit/issues/1298)

SlotSun:
This commit introduced some issues. I suggest we revert it.

PR #1320, opened 2025-11-13
fix: fix flutter338 crash

Predidit:
@alexmercerind 

The Flutter 3.38 stable was released about 12 hours ago, and this release has broken media-kit. The application now crashes whenever the dispose method is called. This occurs in release mode builds on all platforms. This PR fixes the issue. I strongly recommend a new media-kit version release after this PR is merged.

If you are interested in the crash stack, please refer to: https://github.com/Predidit/media-kit/actions/runs/19320211555

### DIFF
--- a/media_kit/lib/src/player/native/core/initializer_native_callable.dart
+++ b/media_kit/lib/src/player/native/core/initializer_native_callable.dart
@@ -60,6 +60,11 @@ class InitializerNativeCallable {
   void dispose(Pointer<generated.mpv_handle> ctx) {
     _locks.remove(ctx.address);
     _eventCallbacks.remove(ctx.address);
+
+    // Clear the wakeup callback in libmpv before closing NativeCallable
+    // to prevent libmpv from invoking a deleted callback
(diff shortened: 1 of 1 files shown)


Predidit:
After further testing, we might be able to delay the release of the new media-kit version. Flutter has obviously messed up quite a few things on Linux in version 3.38. Now, even without hardware acceleration, video frames still experience random rendering errors. Although I still believe the fixes in this PR are necessary, it will fix crash on all platform (android/linux.macos/ios/windows) we should probably wait for a hotfix release from Flutter to ensure our videos render correctly on linux by then.

Predidit:
@alexmercerind 

It seems that Linux is often a lower-priority platform, and I'm concerned that without a high-quality bug report and a minimal reproducible sample, this critical texture widget issue might not get the attention it deserves for a long time.

Unfortunately, I don't have the required expertise in this area. This task seems to require a seasoned Flutter plugin developer with a deep understanding of graphics APIs. Given your extensive experience, I was hoping you might have some time to create a sample and submit it to the official Flutter repository.

PR #1325, opened 2025-11-15
fix(linux): fix h/w render

Predidit:
Our GDK-based context management for the Linux renderer is broken because Flutter now uses its own EGL context instead of GDK's. This corrupts the entire UI with flickering and artifacts when hardware acceleration is enabled, and even causes crashes on some machines.

This PR fixes it by re-implementing context management using EGL and EGLImage for GPU-only texture sharing. A GPU-GPU copy is needed to avoid flickering caused by a suspected Flutter engine issue. The new implementation has been tested on Flutter 3.35.7 and 3.38.0.

I urge a new release after this merge, as the media-kit hardware acceleration on Linux is completely non-functional and unstable right now.

ref: https://github.com/flutter/flutter/commit/407c051ec9eac79b97657ade60d02b7c8e3e0715

### DIFF
--- a/media_kit_video/linux/texture_gl.cc
+++ b/media_kit_video/linux/texture_gl.cc
@@ -250,9 +250,8 @@ gboolean texture_gl_populate_texture(FlTextureGL* texture,
     // Unbind FBO
     glBindFramebuffer(GL_FRAMEBUFFER, 0);
     
-    // Flush and synchronize to ensure rendering is complete
+    // Flush to ensure rendering is complete
     glFlush();
-    glFinish();
     
     // Restore Flutter's context
     eglMakeCurrent(flutter_display, flutter_draw, flutter_read, flutter_context);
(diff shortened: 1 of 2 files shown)


Predidit:
@alexmercerind 

I'm not an expert in this field, so any feedback or questions are welcome.

alexmercerind (approved):
I hope GLES is able to have full non-copy hwdec.

Predidit:
@alexmercerind 

Yes, this is fully hardware-accelerated, and we even have full passthrough for VA-API.

The only issue is that I've been unable to create an EGL context correctly under X11, so the current hardware acceleration only works on Wayland. As a side note, if you are debugging with VS Code, the application will not have hardware acceleration even on a Wayland session, because VS Code itself runs on XWayland.

Also, since we are now using an EGL context instead of a GDK one, a dedicated rendering thread—similar to our Windows implementation—is now possible. This is a major refactor, which I have already implemented and am currently battle-testing in my own open-source project.

I was planning to submit it after more testing, but if you think we should proceed with this implementation now, I can submit a PR later today.

Issue #1345, opened 2025-12-13
[Linux] video tearing when building with flutter 3.38.x


NBA2K1:
I'm using the latest media-kit version. 
When I run with flutter 3.38.x, the video shows tearing.
When I run with 3.35.7, everything is fine.

```
System:
  Kernel: 6.14.0-36-generic arch: x86_64 bits: 64 compiler: gcc v: 13.3.0 clocksource: tsc
  Desktop: Cinnamon v: 6.4.8 tk: GTK v: 3.24.41 wm: Muffin v: 6.4.1 vt: 7 dm: LightDM v: 1.30.0
    Distro: Linux Mint 22.2 Zara base: Ubuntu 24.04 noble
CPU:
  Info: 6-core model: AMD Ryzen 5 5600 bits: 64 type: MT MCP smt: enabled arch: Zen 3+ rev: 2
    cache: L1: 384 KiB L2: 3 MiB L3: 32 MiB
  Speed (MHz): avg: 3293 high: 3845 min/max: 550/4470 boost: enabled cores: 1: 3337 2: 3226
    3: 1750 4: 3297 5: 3845 6: 3595 7: 3759 8: 3637 9: 3147 10: 3189 11: 3583 12: 3157
    bogomips: 84000
  Flags: avx avx2 ht lm nx pae sse sse2 sse3 sse4_1 sse4_2 sse4a ssse3 svm
Graphics:
  Device-1: AMD Ellesmere [Radeon RX 470/480/570/570X/580/580X/590] vendor: Tul / PowerColor
    driver: amdgpu v: kernel arch: GCN-4 pcie: speed: 8 GT/s lanes: 16 ports: active: HDMI-A-1
    empty: DP-1, DP-2, DP-3, DVI-D-1 bus-ID: 29:00.0 chip-ID: 1002:67df class-ID: 0300 temp: 55.0 C
  Display: x11 server: X.Org v: 21.1.11 with: Xwayland v: 23.2.6 driver: X: loaded: amdgpu
    unloaded: fbdev,modesetting,vesa dri: radeonsi gpu: amdgpu display-ID: :0 screens: 1
 
```

When I switch to the experimental Wayland session of Cinnamon, video plays fine.

No change when switching hardware acceleration on/off.

BobMervell:
I have the same issue here with flutter 3.41.3 and mediakit: 
```
  media_kit: ^1.2.6
  media_kit_video: ^2.0.1
[...]

Issue #1287, opened 2025-10-07
Crash on Ubuntu Frame (Wayland / EGL) when disposing video — epoxy_get_proc_address assertion failure


hauketoenjes:
When running a Flutter app using media_kit on Ubuntu Frame (Wayland / Mir, EGL rendering), the app occasionally crashes when closing or disposing a video player. The crash happens inside libepoxy due to a missing current EGL/GLX context. It occurs intermittently—usually when the VideoController widget is removed or disposed.

### Environment:
- Host OS: Ubuntu Core 24
- Snap base: core22
- Display server: Ubuntu Frame (Wayland / Mir, EGL)
- Flutter: stable (3.35.1)
- media_kit: latest release (1.2.1)
- Hardware: Embedded x86 device without dedicated GPU (Mesa / DRI driver)

### Logs
```
Cannot load libcuda.so.1
media_kit: VideoOutput: Using H/W rendering.
VideoOutput.Resize
{handle: 123231404752048, id: 97965319568032, rect: {left: 0, top: 0, width: 3840, height: 2160}}
../src/dispatch_common.c:872: epoxy_get_proc_address: Assertion `0 && "Couldn't find current GLX or EGL context.\n"' failed.
Aborted
```

### What I’ve tried
Set environment variables in the systemd service:

```ini
Environment="MPV_GPU_CONTEXT=wayland"
Environment="MPV_OPENGL_ES=yes"
Environment="LIBGL_DRIVER=dri"
```

Crash still occurs occasionally on dispose.

### Expected Behavior

Video playback should stop and dispose gracefully without OpenGL/EGL context assertions.

### Actual Behavior

When disposing the VideoController, libepoxy fails to find a valid current EGL or GLX context:

```
epoxy_get_proc_address: Assertion `0 && "Couldn't find current GLX or EGL context.\n"' failed.
Aborted
```

### Possible Cause

It seems mpv’s EGL surface is destroyed while media_kit still holds an active render callback, or the teardown sequence calls into GL after context loss.
Likely specific to Wayland/Mir embedded environments where no X11/GLX context exists.

### Questions
- Is there a recommended disposal sequence for Wayland/EGL environments?
- Should media_kit automatically enforce gpu-context=wayland and opengl-es=yes when running under Ubuntu Frame?
- Any known issues with libepoxy teardown on Wayland-only systems?

### Relevant Code
<details>
<summary>Example on how i use the player in my project</summary>

```dart
[...]

PR #1346, opened 2025-12-14
feat(linux): dedicated render thread

Predidit:
This is a brand-new Linux renderer featuring an independent rendering thread, modeled after our current Windows implementation. On the Linux platform, the OpenGL driver's implicit synchronization is highly unreliable when it comes to cross-thread texture sharing. We cannot rely on simple `glFinish` for forced synchronization because it would cause severe rendering pipeline blocking and performance degradation, which contradicts the original purpose of implementing an independent rendering thread. Therefore, we had to adopt a complex triple-buffering implementation to eliminate most of the implicit behavior, inspired by the Vulkan swap chain. This implementation is highly optimized and has been battle-tested in my open-source project—I can confirm there are no crashes or memory leaks.

There is still room for further optimization, similar to our current Windows implementation. When multiple player instances are running, they share the same rendering thread, which can cause stuttering. I am currently working on a prototype where each player instance has its own independent rendering process. I will submit another PR once it is completed and thoroughly tested.


media_kit_video/linux/video_output_manager.cc
```diff
@@ -13,18 +13,21 @@ struct _VideoOutputManager {
...
   FlView* view;
+  GLRenderThread* gl_render_thread;
 };
 
 G_DEFINE_TYPE(VideoOutputManager, video_output_manager, G_TYPE_OBJECT)
 
 static void video_output_manager_init(VideoOutputManager* self) {
   self->video_outputs = g_hash_table_new_full(g_direct_hash, g_direct_equal,
                                               nullptr, g_object_unref);
+  self->gl_render_thread = new GLRenderThread();  // Dedicated GL render thread
```
alexmercerind: Lets use `g_autoptr` (similar to VideoOutputManager itself). It's a convention to use std:: smart pointers on Windows & GLib on Windows in the project.

media_kit_video/linux/video_output_manager.cc
```diff
@@ -13,18 +13,21 @@ struct _VideoOutputManager {
...
 }
 
 static void video_output_manager_dispose(GObject* object) {
   VideoOutputManager* self = VIDEO_OUTPUT_MANAGER(object);
   g_hash_table_unref(self->video_outputs);
+  delete self->gl_render_thread;
```
alexmercerind: Can be removed once above is solved.

media_kit_video/linux/gl_render_thread.cc
```diff
@@ -0,0 +1,103 @@
+// This file is a part of media_kit
+// (https://github.com/media-kit/media-kit).
+//
+// Copyright © 2025 Predidit.
+// All rights reserved.
+// Use of this source code is governed by MIT license that can be found in the
+// LICENSE file.
+
+#include "include/media_kit_video/gl_render_thread.h"
```
alexmercerind: The implementation seems very identical to ThreadPool used on Windows.
Please check if it can be shared (common folder), if not, then make it a GTK/GLib API.

media_kit_video/linux/texture_gl.cc
```diff
@@ -6,16 +6,31 @@
 // Use of this source code is governed by MIT license that can be found in the
 // LICENSE file.
 
+// The file is moded for the purpose of triple buffering using mailbox model.
+// Copyright © 2025 Predidit
+// All rights reserved.
```
alexmercerind: Same.

media_kit_video/linux/video_output.cc
```diff
@@ -6,9 +6,14 @@
 // Use of this source code is governed by MIT license that can be found in the
 // LICENSE file.
 
+// The file is moded for the purpose of triple buffering using mailbox model.
+// Copyright © 2025 Predidit
+// All rights reserved.
```
alexmercerind: Same.

Predidit:
@alexmercerind 

Hi Alex,

Sorry for the ping, but I’m following up on this PR as it has been open for quite some time. Could we take a look at reviewing this?

This implementation has undergone extensive testing in my fork and another open-source project, proving to be highly stable. It offers significant performance gains and resolves several long-standing Linux legacy issues, including but not limited to #1404, #1345, and #1287.

khlsvr:
@Carapacik  I tested the renderer from this PR in a real application, and unfortunately it does not fix the X11 stuttering I'm seeing.

Some context: Mangatan and Mangayomi are two closely related Flutter-based manga/anime applications that use media_kit for video playback. My original report is kodjodevf/mangayomi#974.

My setup is Void Linux + X.Org/X11, Ryzen 5 4500U/Radeon Renoir, Mesa 26.1.8. The same anime scene intermittently stutters under X11, while it is generally smooth under Sway/Wayland. XLibre also behaves the same as X.Org.

I tested the Mangatan beta release that incorporates this renderer (so I tested the renderer through the application rather than building media_kit myself https://github.com/1Selxo/Mangatan/commit/735865e955afd764c474e75025ea0b0e1760d691). Unfortunately, the same scene still stutters.

The console in the newer Mangatan build still reports:

`media_kit: VideoOutput: Failed to get Flutter's EGL display or context.`
`media_kit: VideoOutput: S/W rendering.`

I also tested hardware decoding enabled/Auto and disabled/software decoding. The stutter remains. With hardware decoding enabled, libmpv reports `vulkan-copy` / `nv12`, but the presentation still stutters.

The most interesting A/B result is with the related Mangayomi application:

* Newer Mangayomi → stutters
* Mangatan → stutters
* Mangayomi 0.6.75 → the exact same scene plays smoothly on the same X11 system

The old Mangayomi 0.6.75 also reports:

`media_kit: VideoOutput: Using S/W rendering.`

So S/W rendering itself does not seem to explain the problem.

There is also an interesting historical clue from NBA2K1's comment on https://github.com/kodjodevf/mangayomi/issues/974 : they reported that the problem was present since Mangayomi v0.6.75 but not v0.6.70, and pointed to media-kit #1345. That issue reports X11 rendering problems with Flutter 3.38.x while Flutter 3.35.7 was fine and Wayland worked.

Taken together, this makes me suspect that the problem is somewhere in the newer Flutter/Linux/X11 presentation path rather than simply video decoding. I can't say that Flutter version is definitively the cause, but the old Mangayomi build being smooth on the same machine is a fairly strong A/B result.

For clarity, the renderer in this PR was tested through Mangatan, while the old/new Flutter comparison comes from Mangayomi. I'm using them as related test cases, not treating them as the same application.

Edit: Reading this post 9h later I notice that I tested the 0.6.75 version of mangayomi without the stutter issue, while that should be the version that first has the problem. I went now to download the appimage of the mentioned last working release of 0.6.70, but it seems that there are currently some issues with the streaming provider. I will be updating this post later on which mangayomi version is the newest that would give me no stutter.

Predidit:
@khlsvr 

This PR does not include X11 support, it currently only implements hardware acceleration on Wayland. When I originally opened this PR, I assumed few people still cared about X11, and the Flutter engine itself suffers from annoying readback overhead on X11, so that code path was left unimplemented.

That said, I now realize that having an X11 fallback is still worthwhile. In the meantime, feel free to test my branch at https://github.com/Predidit/media-kit , which includes X11 support. The reason I haven't updated this PR directly is that the diff is already quite large. Once this gets merged, I will open a separate follow up PR for X11 compatibility.

Issue #964, opened 2024-09-27
Multiple videos not playing correctly


wrngwrld:
Whenever I try to run the example [09.seamless.dart](https://github.com/media-kit/media-kit/blob/main/media_kit_test/lib/tests/09.seamless.dart) and set the AudioTrack to auto instead of none. I always get this stutter on page changes/the videos are not playing. Is there a fix for this? It probably is a problem of the dispose method, as disabling the dispose in this example does make it run smoother.

https://github.com/user-attachments/assets/1dc79314-26b7-42a4-91f5-028a7e1b6747

alexmercerind:
Yeah... this is a thing with iOS. I haven't figured out how to allow multiple audio sources to run in parallel.

Be assured that this is not stutter... but iOS system pauses another audio as soon as one one starts playing.

wrngwrld:
I'm actually not too sure, that it is iOS pausing the audio rather than something weird happening in the dispose when the audio is enabled. When I removed these lines (https://github.com/media-kit/media-kit/blob/264a54f23e6e0b0c564800de39f547b074f1b28a/media_kit_test/lib/tests/09.seamless.dart#L89-L103), the players no longer dispose, and the lag between them disappears as well.

alexmercerind:
With audio enabled?

wrngwrld:
Yes, with the audio enabled

alexmercerind:
Interesting...

alexmercerind:
Let us know if you find something... I don't own an iOS, try to comment out things inside dispose method maybe.

wrngwrld:
The weird thing is that the only thing, that fixes it is by commenting out this line https://github.com/media-kit/media-kit/blob/264a54f23e6e0b0c564800de39f547b074f1b28a/media_kit_test/lib/tests/09.seamless.dart#L93

So there probably is some code in the platform dispose method, that makes it lag. I'll try to take a look into that code when I have time :)

alexmercerind:
I don't see the same on any other platform... I still doubt its an audio issue. We do modify --aid etc. in dispose

alexmercerind:
Try building against main branch now.

wrngwrld:
Still no improvement :(

doersoul:
Calling player.dispose(), will trigger the suspension of the audio session, causing other players to also pause playing

solution

```dart
  /// Disposes the [Player] instance & releases the resources.
  @override
  Future<void> dispose({bool synchronized = true}) {
[...]

PR #1352, opened 2025-12-20
fix: fix flutter338 hot restart crash

### LINKED ISSUE #1340: Crash on Hot Restart in Flutter 3.38.x on Android and iOS
Hi,
I encountered a crash in `media_kit` during **Hot Restart** in Flutter 3.38.4. The issue occurs on both physical devices (Android and iOS) and simulators/emulators.

This crash **does not occur** on Flutter 3.35.x.

---

**Error details on Android:**

```
Performing hot restart...
Syncing files to device SM A525F...
E/Dart    ( 6630): ../../../flutter/third_party/dart/runtime/vm/runtime_entry.cc: 5113: error: Callback invoked after it has been deleted.
E/Dart    ( 6630): ../../../flutter/third_party/dart/runtime/vm/runtime_entry.cc: 5113: error: Callback invoked after it has been deleted.
E/DartVM  ( 6630): version=3.10.3 (stable) (Tue Dec 2 01:04:53 2025 -0800) on "android_arm64"
E/DartVM  ( 6630): pid=6630, thread=508284091568, isolate_group=(nil)(0x0), isolate=(nil)(0x0)
E/DartVM  ( 6630): os=android, arch=arm64, comp=yes, sim=no
E/DartVM  ( 6630): isolate_instructions=0, vm_instructions=76970401a0
E/DartVM  ( 6630): fp=765817b630, sp=765817b600, pc=7696e47fd4
E/DartVM  ( 6630):   pc 0x0000007696e47fd4 fp 0x000000765817b630 /data/app/~~L4MUk4qri2WR65Lx_VpIQA==/com.example.example-xV3laKXo0bsK9RLRiO15dA==/base.apk!/lib/arm64-v8a/libflutter.so+0x1ff9fd4
E/DartVM  ( 6630): Aborting reentrant request for stack trace.
E/DartVM  ( 6630):   pc 0x0000007696d35ec0 fp 0x000000765817b7c8 /data/app/~~L4MUk4qri2WR65Lx_VpIQA==/com.example.example-xV3laKXo0bsK9RLRiO15dA==/base.apk!/lib/arm64-v8a/libflutter.so+0x1ee7ec0
E/DartVM  ( 6630):   pc 0x000000766e6e3bf4 fp 0x000000765817b8b0 /data/app/~~L4MUk4qri2WR65Lx_VpIQA==/com.example.example-xV3laKXo0bsK9RLRiO15dA==/base.apk!/lib/arm64-v8a/libmpv.so+0x406bf4
E/DartVM  ( 6630):   pc 0x000000766e6e3bf4 fp 0x000000765817b950 /data/app/~~L4MUk4qri2WR65Lx_VpIQA==/com.example.example-xV3laKXo0bsK9RLRiO15dA==/base.apk!/lib/arm64-v8a/libmpv.so+0x406bf4
E/DartVM  ( 6630):   pc 0x000000766e700278 fp 0x000000765817b9b0 /data/app/~~L4MUk4qri2WR65Lx_VpIQA==/com.example.example-xV3laKXo0bsK9RLRiO15dA==/base.apk!/lib/arm64-v8a/libmpv.so+0x423278
E/DartVM  ( 6630):   pc 0x000000766e7025fc fp 0x000000765817ba00 /data/app/~~L4MUk4qri2WR65Lx_VpIQA==/com.example.example-xV3laKXo0bsK9RLRiO15dA==/base.apk!/lib/arm64-v8a/libmpv.so+0x4255fc
E/DartVM  ( 6630):   pc 0x000000766e6fc454 fp 0x000000765817bba0 /data/app/~~L4MUk4qri2WR65Lx_VpIQA==/com.example.example-xV3laKXo0bsK9RLRiO15dA==/base.apk!/lib/arm64-v8a/libmpv.so+0x41f454
E/DartVM  ( 6630):   pc 0x000000766e6fb038 fp 0x000000765817bc00 /data/app/~~L4MUk4qri2WR65Lx_VpIQA==/com.example.example-xV3laKXo0bsK9RLRiO15dA==/base.apk!/lib/arm64-v8a/libmpv.so+0x41e038
E/DartVM  ( 6630):   pc 0x000000766e6e15bc fp 0x000000765817bc30 /data/app/~~L4MUk4qri2WR65Lx_VpIQA==/com.example.example-xV3laKXo0bsK9RLRiO15dA==/base.apk!/lib/arm64-v8a/libmpv.so+0x4045bc
E/DartVM  ( 6630):   pc 0x00000077bc50d0f8 fp 0x000000765817bc50 /apex/com.android.runtime/lib64/bionic/libc.so+0xfd0f8
E/DartVM  ( 6630):   pc 0x00000077bc4a6a08 fp 0x000000765817bc80 /apex/com.android.runtime/lib64/bionic/libc.so+0x96a08
E/DartVM  ( 6630): -- End of DumpStackTrace
F/libc    ( 6630): Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 6977 (Thread-7), pid 6630 (example.example)
Restarted application in 2.794ms.
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
Build fingerprint: 'samsung/a52qnseea/a52q:14/UP1A.231005.007/A525FXXSCFYF1:user/release-keys'
Revision: '8'
ABI: 'arm64'
Processor: '6'
Timestamp: 2025-12-05 00:25:41.659853010+0100
Process uptime: 38s
Cmdline: com.example.example
pid: 6630, tid: 6977, name: Thread-7  >>> com.example.example <<<
uid: 10675
signal 6 (SIGABRT), code -1 (SI_QUEUE), fault addr --------
Abort message: '../../../flutter/third_party/dart/runtime/vm/runtime_entry.cc: 5113: error: Callback invoked after it has been deleted.'
    x0  0000000000000000  x1  0000000000001b41  x2  0000000000000006  x3  0000007649f60590
    x4  0000000000000000  x5  0000000000000000  x6  0000000000000000  x7  0000000011c8ccec
    x8  00000000000000f0  x9  00000077bc437f28  x10 0000000000000001  x11 00000077bc4b4344
    x12 0000007649f5ee10  x13 0000000000000000  x14 0000007649f5eec0  x15 0000007649f607c0
    x16 00000077bc51fa00  x17 00000077bc4f7e60  x18 0000007649936000  x19 00000000000019e6
    x20 0000000000001b41  x21 00000000ffffffff  x22 0000007649f607d8  x23 0000007649f607e0
    x24 b4000076a955c7d0  x25 0000000000000000  x26 b4000076a9402850  x27 0000000000000000
    x28 000000766ee39e58  x29 0000007649f60610
    lr  00000077bc4a5084  sp  0000007649f60570  pc  00000077bc4a50b0  pst 0000000000001000
4 total frames
backtrace:
      #00 pc 00000000000950b0  /apex/com.android.runtime/lib64/bionic/libc.so (abort+168) (BuildId: 02a91a85343debb2911714273ff2b670)
      #01 pc 0000000001ee7ec4  /data/app/~~L4MUk4qri2WR65Lx_VpIQA==/com.example.example-xV3laKXo0bsK9RLRiO15dA==/base.apk!libflutter.so (offset 0x12e8000) (BuildId: f68b450bf0ae79945577495c94ad9af56b81e07e)
      #02 pc 0000000002012418  /data/app/~~L4MUk4qri2WR65Lx_VpIQA==/com.example.example-xV3laKXo0bsK9RLRiO15dA==/base.apk!libflutter.so (offset 0x12e8000) (BuildId: f68b450bf0ae79945577495c94ad9af56b81e07e)
      #03 pc 000000000001003c  [anon:FfiCallbackMetadata::TrampolinePage]
Lost connection to device.

...
```

In short, it seems a **callback is invoked after being deleted**, causing the Dart VM to crash.

---

**Used packages:**

```yaml
media_kit: ^1.2.3
media_kit_video: ^2.0.1
media_kit_libs_video: ^1.0.7
```

---

**Flutter version:**

```
Flutter 3.38.4
Dart 3.10.3
```

---

**Steps to reproduce:**

1. Create a Flutter project using `media_kit` + `media_kit_video`.
2. Run the app on an Android or iOS device (physical or simulator).
3. Perform a **Hot Restart**.
4. The crash occurs immediately.

---

**Expected behavior:**
The Hot Restart should complete without crashing.

---

**Observed behavior:**
Crash with error:
`Callback invoked after it has been deleted.`

---

**Request:**
Could you please investigate and release a fix compatible with Flutter 3.38.x?

I can provide additional logs or details if needed.

Thank you very much for your amazing package! I really appreciate all the hard work you put into it.

Best regards,
Leonard

Predidit:
Fixes https://github.com/media-kit/media-kit/issues/1340

The thread merging in the newer version of Flutter forces NativeCallable to be reset during hot restart. There is no clean way for us to receive the hot restart signal and handle it appropriately, making it difficult and error-prone to fix hot restart issues within the NativeCallable framework. I looked into how the flutter_asound plugin handles this, and it appears quite hacky.
This PR chooses to disable NativeCallable in debug mode and fall back to the old Dart isolate–based implementation, allowing us to resolve the issue with minimal cost.

ref: https://github.com/flutter/flutter/issues/178548

### DIFF
--- a/media_kit/lib/src/player/native/core/initializer.dart
+++ b/media_kit/lib/src/player/native/core/initializer.dart
@@ -4,6 +4,7 @@
 /// All rights reserved.
 /// Use of this source code is governed by MIT license that can be found in the LICENSE file.
 import 'dart:ffi';
+import 'dart:isolate';
 
 import 'package:media_kit/generated/libmpv/bindings.dart' as generated;
 import 'package:media_kit/src/player/native/core/execmem_restriction.dart';
@@ -44,7 +45,8 @@ class Initializer {
     // based implementation to avoid native -> Dart callbacks that can outlive
     // the isolate and crash with "Callback invoked after it has been deleted".
     // See: https://github.com/media-kit/media-kit/issues/1340
-    if (kDebugMode) {
+    // We still use NativeCallable based implementation in release mode and unit tests for better performance.
+    if (kDebugMode && isMainIsolate()) {
       return InitializerIsolate().create(callback, options: options);
     }
     if (!isExecmemRestricted) {
@@ -56,7 +58,7 @@ class Initializer {
 
   /// Disposes [Pointer<mpv_handle>].
   void dispose(Pointer<generated.mpv_handle> ctx) {
-    if (kDebugMode) {
+    if (kDebugMode && isMainIsolate()) {
       InitializerIsolate().dispose(mpv, ctx);
       return;
     }
@@ -65,5 +67,10 @@ class Initializer {
     } else {
       InitializerIsolate().dispose(mpv, ctx);
     }
+  }
+
+  bool isMainIsolate() {
+    final name = Isolate.current.debugName;
+    return name == 'main';
   }
 }
--- a/media_kit/lib/src/player/native/core/initializer.dart
+++ b/media_kit/lib/src/player/native/core/initializer.dart
@@ -43,7 +43,7 @@ class Initializer {
     // registered `NativeCallable` trampolines. In debug mode, prefer the isolate
     // based implementation to avoid native -> Dart callbacks that can outlive
     // the isolate and crash with "Callback invoked after it has been deleted".
-    // See: https://github.com/media-kit/media-kit/issues/1348
+    // See: https://github.com/media-kit/media-kit/issues/1340
     if (kDebugMode) {
       return InitializerIsolate().create(callback, options: options);
     }
--- a/media_kit/lib/src/player/native/core/initializer.dart
+++ b/media_kit/lib/src/player/native/core/initializer.dart
@@ -9,6 +9,7 @@ import 'package:media_kit/generated/libmpv/bindings.dart' as generated;
 import 'package:media_kit/src/player/native/core/execmem_restriction.dart';
 import 'package:media_kit/src/player/native/core/initializer_isolate.dart';
 import 'package:media_kit/src/player/native/core/initializer_native_callable.dart';
+import 'package:media_kit/src/values.dart';
 
 /// {@template initializer}
 ///
@@ -38,6 +39,14 @@ class Initializer {
     Future<void> Function(Pointer<generated.mpv_event>) callback, {
     Map<String, String> options = const {},
   }) async {
+    // Hot-restart tears down the Dart isolate, which invalidates any previously
+    // registered `NativeCallable` trampolines. In debug mode, prefer the isolate
+    // based implementation to avoid native -> Dart callbacks that can outlive
+    // the isolate and crash with "Callback invoked after it has been deleted".
+    // See: https://github.com/media-kit/media-kit/issues/1348
+    if (kDebugMode) {
+      return InitializerIsolate().create(callback, options: options);
+    }
     if (!isExecmemRestricted) {
       return InitializerNativeCallable(mpv).create(callback, options: options);
     } else {
@@ -47,6 +56,10 @@ class Initializer {
 
   /// Disposes [Pointer<mpv_handle>].
   void dispose(Pointer<generated.mpv_handle> ctx) {
+    if (kDebugMode) {
+      InitializerIsolate().dispose(mpv, ctx);
+      return;
+    }
     if (!isExecmemRestricted) {
       InitializerNativeCallable(mpv).dispose(ctx);
     } else {
--- a/media_kit/test/src/player/native/core/initializer_test.dart
+++ b/media_kit/test/src/player/native/core/initializer_test.dart
@@ -13,12 +13,44 @@ import 'package:test/test.dart';
 import 'package:media_kit/ffi/ffi.dart';
 
 import 'package:media_kit/src/player/native/core/initializer.dart';
+import 'package:media_kit/src/player/native/core/initializer_isolate.dart';
+import 'package:media_kit/src/player/native/core/initializer_native_callable.dart';
 import 'package:media_kit/src/player/native/core/native_library.dart';
+import 'package:media_kit/src/player/native/core/execmem_restriction.dart';
+import 'package:media_kit/src/values.dart';
 
 import 'package:media_kit/generated/libmpv/bindings.dart';
 
 MPV? _mpv;
 MPV get mpv => _mpv!;
+
+Future<Pointer<mpv_handle>> _createProductionLikeHandle(
+  Future<void> Function(Pointer<mpv_event>) callback, {
+  Map<String, String> options = const {},
+}) {
+  // `Initializer` intentionally forces `InitializerIsolate` in debug mode to
+  // avoid hot-restart related issues with `NativeCallable` trampolines.
+  // These unit tests, however, want to exercise production behavior.
+  if (kDebugMode) {
+    if (!isExecmemRestricted) {
+      return InitializerNativeCallable(mpv).create(callback, options: options);
+    }
+    return InitializerIsolate().create(callback, options: options);
+  }
+  return Initializer(mpv).create(callback, options: options);
+}
+
+void _disposeProductionLikeHandle(Pointer<mpv_handle> handle) {
+  if (kDebugMode) {
+    if (!isExecmemRestricted) {
+      InitializerNativeCallable(mpv).dispose(handle);
+    } else {
+      InitializerIsolate().dispose(mpv, handle);
+    }
+    return;
+  }
+  Initializer(mpv).dispose(handle);
+}
 
 void main() {
   setUp(() {
@@ -29,7 +61,7 @@ void main() {
     'initializer-init',
     () {
       expect(
-        Initializer(mpv).create((_) async {}),
+        _createProductionLikeHandle((_) async {}),
         completes,
       );
     },
@@ -38,7 +70,7 @@ void main() {
     'initializer-create',
     () {
       expect(
-        Initializer(mpv).create((_) async {}),
+        _createProductionLikeHandle((_) async {}),
         completes,
       );
     },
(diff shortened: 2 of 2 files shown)


Predidit:
I've just noticed that some tests are broken, primarily due to ambiguities in test implementation. This issue occurs when running NativeCallable tests after creating instances using Dart isolates. I'll promptly fix this to ensure the tests clearly and consistently use NativeCallable, aligning with the actual production environment.

Predidit:
Broken initializer_test.dart has been fixed.

Predidit:
@alexmercerind 

Merry Christmas! I believe this PR has reached a local optimum. In the latest commit, I’ve implemented a slightly "hacky" workaround: detecting the current isolate name to distinguish between unit tests and running the full app in debug mode. We now only trigger the rollback when the full app is running. This ensures our unit test scenarios remain identical to production without needing to refactor the player initialization logic in tests. Could you take a look at the current implementation? Unless we find an elegant way to handle hot-restart signal, I think this approach is solid.

alexmercerind:
I always wanted to delete isolate based initializer, but someone brought it back in #1022. It is now only used in adverse situations. The main problem with isolate based initializer is that we can only create only a limited number of concurrent instances & after a limit (7-8; based on number of hardware threads in CPU), the whole app freezes.

Predidit:
Based on the comments in flutter/flutter#178548, we have essentially hit an impasse. Our `NativeCallable` solution is elegant, high-performance, and cross-platform. However, the Flutter team maintains that the previous behavior—where a Use-After-Free (UAF) resulted in a no-op instead of a crash—was improper. Since the new NativeCallable implementation mandates a VM crash on UAF, and Flutter currently provides no "hot-restart" signal for us to cancel pending callbacks, UAF becomes almost inevitable, leading to crashes during development.

I believe this is a unique constraint of Dart isolates. Perhaps we should simply log a message informing users that experiencing app freezes in debug mode when holding more than 8 players is "expected behavior" given these platform limitations.

Aside from that, our only other potential workaround would be switching to MethodChannel, which would require writing platform-specific glue code for every target. This, however, would be a complete departure from the core design philosophy of media-kit.

icecode-ai:
I’ve modified the dispose method. It works fine on both Android and iOS, but I’m not sure if this implementation might have any side effects

```dart
  /// Disposes the [Player] instance & releases the resources.
  @override
  Future<void> dispose({bool synchronized = true}) {
    Future<void> function() async {
      if (disposed) {
        throw AssertionError('[Player] has been disposed');
      }

      await waitForPlayerInitialization;
      await waitForVideoControllerInitializationIfAttached;

      await NativeReferenceHolder.instance.remove(ctx);

      // tips: modified here
      if (Platform.isIOS) {
        await _command(['quit']);
      }

      await stop(notify: false, synchronized: false);

      disposed = true;

      await super.dispose();

      Initializer(mpv).dispose(ctx);

      // tips: modified here
      if (!Platform.isIOS) {
        Future.delayed(const Duration(seconds: 5), () {
          _command(['quit']);

          mpv.mpv_terminate_destroy(ctx);
        });
      }
    }

    if (synchronized) {
      return lock.synchronized(function);
    } else {
      return function();
    }
  }
```

alexmercerind:
@doersoul

Can you send a PR? We'll be happy to merge if all tests pass.

Also why only `Platform.isIOS`? Can't we keep same for all platforms?

Predidit:
I haven't explored the specific flow of hot-restart yet, but if this draft works, I wouldn’t be surprised at all — 5 seconds is plenty of time to complete a hot restart. Personally, I don’t think delaying the dispose call by 5 seconds is a good idea. However, I trust Alex’s professional judgment on this. I have more devices available for testing and will try out this implementation on Linux and Windows tomorrow.

alexmercerind:
Haha... nothing professional about me. Just scared to touch things haha.

icecode-ai:
I don't know the specific root cause, and I'm not a professional. These findings are based on my tests regarding different behaviors on Android vs. iOS when disposing of a player instance among several others [#964](https://github.com/media-kit/media-kit/issues/964)

Predidit:
On non-iOS platforms, this draft executes `_command(['quit'])` before calling `mpv.mpv_terminate_destroy(ctx)`. However, this approach does not resolve the hot-reload crashes on Windows and Linux. To be honest, this implementation is a bit unusual—we are essentially performing an asynchronous cleanup via quit followed immediately by a synchronous terminate_destroy, resulting in a "double cleanup" without waiting for the async signal. That said, it provides some useful inspiration; I will explore whether a full asynchronous cleanup approach can fix the hot-reload issues.

ItsThatBlackKid:
Is this still being reviewed?

birros (approved):
See the conversation in https://github.com/media-kit/media-kit/pull/1392

ashkryab-grnd:
I can reproduce this in a production release build on Android, without using hot restart.

Environment:

- Flutter 3.44.0 / Dart 3.12.0
- `media_kit` commit `2e35dbde9fd10bde0a63efa83868c07f5898ffb5`
- A looping background video is disposed when navigating away from the screen

A few seconds after `Player.dispose()`, the application crashes on an unrelated screen with:

```text
Callback invoked after it has been deleted.
Fatal signal SIGABRT (6)
[anon:FfiCallbackMetadata::TrampolinePage]
```

This appears to be the production disposal race described in this PR. The current order closes the `NativeCallable` before the delayed `mpv_terminate_destroy(ctx)`, leaving a window where the mpv core thread can invoke the deleted callback.

The cleanup order should be changed so that the native callable remains alive until the mpv context and its core thread are fully terminated:

```text
mpv_terminate_destroy(ctx)
NativeCallable.close()
```

PR #1352 does not cover this case because its isolate fallback is enabled only in debug mode. It would be useful to reopen or reconsider this fix for release builds.

PR #1375, opened 2026-01-27
refactor: add 'keep-open' flag to PlayerConfiguration

gaganyadav80:
- Using this we can control the auto-advancing behaviour of playlist in `mpv`

### DIFF
--- a/media_kit/lib/src/player/native/player/real.dart
+++ b/media_kit/lib/src/player/native/player/real.dart
@@ -8,23 +8,18 @@ import 'dart:collection';
 import 'dart:ffi';
 import 'dart:io';
 import 'dart:typed_data';
+
 import 'package:image/image.dart';
-import 'package:meta/meta.dart';
-import 'package:path/path.dart';
-import 'package:safe_local_storage/safe_local_storage.dart';
-import 'package:synchronized/synchronized.dart';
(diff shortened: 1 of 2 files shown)


gaganyadav80:
I will raise another PR which will allow passing a Map<String, String> to set custom properties. Which I guess is better if someone wants to add multiple custom properties or override the defaults ones.

alexmercerind:
> I will raise another PR which will allow passing a Map<String, String> to set custom properties. Which I guess is better if someone wants to add multiple custom properties or override the defaults ones.

This will not be merged, because it is incorrect. 3 reasons:

- You are exposing backend (mpv) specific properties into common `Player`/`PlayerConfiguration`. I know there are few other properties like this e.g. `logLevel` which already exist, but mistakes have been made.
- Modifying internal behavior will cause issues, the state/stream will break if you set --keep-open=no. We as a project, cannot support everything & its dependent's own responsibility if they rely on internal/undocumented (from media_kit POV) APIs.
- Most importantly, an option to achieve this already exists (no support; read comment on the method)! Refer:

https://github.com/media-kit/media-kit/issues/14#issuecomment-1373205132

gaganyadav80:
@alexmercerind  Thanks for the explanation and for pointing me to the existing option. Appreciate you taking the time to clarify.

PR #1386, opened 2026-02-17
Fix SIGABRT on hot restart by clearing mpv wakeup callbacks before quit

### LINKED ISSUE #1314: Crash during hot restart of the application [flutter 3.38+]
When using media_kit, the application crashes when attempting to use hot restart on the beta/master channels of flutter with the following log:

```
../../../flutter/third_party/dart/runtime/vm/runtime_entry.cc: 5113: error: Callback invoked after it has been deleted.
version=3.10.0-290.1.beta (beta) (Tue Oct 7 02:58:29 2025 -0700) on “windows_x64”
pid=19684, thread=25120, isolate_group= (nil)(0000000000000000), isolate=(nil)(0000000000000000)
os=windows, arch=x64, comp=no, sim=no
isolate_instructions=0, vm_instructions=7ffda640e280
fp=93b98ff560, sp=93b98ff518, pc=7ffda61a7703
  pc 0x00007ffda61a7703 fp 0x00000093b98ff560 InternalFlutterGpu_Texture_AsImage+0x14e16f
-- End of DumpStackTrace
Lost connection to device.

Exited.
```

The platform is most likely irrelevant (tested on Windows and Android). Reproduced with code from https://github.com/media-kit/media-kit/blob/main/media_kit/example/README.md

jmgeffroy:
On Flutter 3.38+, hot restart destroys Dart NativeCallable trampolines before the native mpv thread stops. NativeReferenceHolder sends mpv_command_string("quit") to old handles, which triggers mpv event processing and invokes the dead wakeup callback → SIGABRT.

Fix: call mpv_set_wakeup_callback(handle, nullptr, nullptr) on each old handle before sending quit. This is synchronous — once it returns, mpv will never invoke the dead trampoline again.

Fixes: https://github.com/media-kit/media-kit/issues/1314

### DIFF
--- a/media_kit/lib/src/player/native/player/real.dart
+++ b/media_kit/lib/src/player/native/player/real.dart
@@ -51,8 +51,17 @@ void nativeEnsureInitialized({String? libmpv}) {
     print('$tag Found ${references.length} reference(s).');
     print('$tag Disposing:\n${references.map((e) => e.address).join('\n')}');
 
-    // I can only get quit to work; [mpv_terminate_destroy] causes direct crash.
+    // First, clear wakeup callbacks on all old handles to prevent mpv from
+    // invoking deleted Dart NativeCallable trampolines (SIGABRT on Flutter 3.38+).
+    // mpv_set_wakeup_callback is synchronous: once it returns, mpv will never
+    // call the old (dead) trampoline again.
+    // See: https://github.com/media-kit/media-kit/issues/1314
     final mpv = generated.MPV(DynamicLibrary.open(NativeLibrary.path));
+    for (final reference in references) {
+      mpv.mpv_set_wakeup_callback(reference.cast(), nullptr, nullptr);
+    }
+
+    // Now it's safe to send quit; mpv won't try to notify Dart anymore.
     final cmd = 'quit'.toNativeUtf8();
     try {
       for (final reference in references) {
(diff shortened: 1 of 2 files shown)


Predidit:
Thanks for your work, this patch works very well in my local testing.

I'm not sure what @alexmercerind will think about it, though. The change isn't entirely clean from an architectural perspective, since it breaks the abstraction boundary and callable objects are no longer managed in a single place.

Of course, that's not your fault. We're essentially working around limitations on the Flutter side here.

alexmercerind:
Hi @jmgeffroy,

The fix looks clean, can you revert the last commit in this branch however?

A more universal API to achieve that already exists in the package:

https://github.com/media-kit/media-kit/issues/14#issuecomment-1373205132

birros:
@jmgeffroy Hello! Could you please revert commit dafe04e8fa7cbda17753140880b00735fda7a905 as requested by alexmercerind?

Alternatively, enabling "Allow edits from maintainers" on your PR would let us take care of it directly.

If neither works for you, we might need to open a new PR based on your changes, we'd of course credit you as the original author. Let us know what works best! Thx

birros:
Closing this PR in favor of https://github.com/media-kit/media-kit/pull/1416, which continues this work.

Thanks @jmgeffroy for the original contribution!

PR #1392, opened 2026-03-14
fix: resolve hot restart crash (SIGABRT) on Flutter 3.38+

### LINKED ISSUE #1314: Crash during hot restart of the application [flutter 3.38+]

### LINKED ISSUE #1340: Crash on Hot Restart in Flutter 3.38.x on Android and iOS

realmeylisdev:
## Summary

Fixes the **SIGABRT crash on hot restart** affecting all platforms with Flutter 3.38+ / Dart 3.10+.

- Clears mpv wakeup callbacks on stale handles before sending `quit` during NativeReferenceHolder cleanup
- Falls back to isolate-based event loop in debug mode to avoid NativeCallable trampolines entirely
- Adds defensive null-checks in the NativeCallable callback handler

## Root Cause

Flutter 3.38 merged the UI/Platform threads and changed `NativeCallable` lifecycle: invoking a dead trampoline now crashes with SIGABRT instead of being a silent no-op. During hot restart:

1. Dart destroys all `NativeCallable` trampolines
2. Native mpv threads keep running and invoke the dead wakeup callback
3. Dart VM detects use-after-free → `SIGABRT`

The Flutter team [confirms this is intentional](https://github.com/flutter/flutter/issues/178548) and provides no hot-restart notification API ([#75528](https://github.com/flutter/flutter/issues/75528) open since 2021).

### Layer 1: Clear wakeup callbacks in NativeReferenceHolder cleanup (`real.dart`)

Before sending `quit` to stale handles after hot restart, call `mpv_set_wakeup_callback(handle, nullptr, nullptr)`. This is **synchronous** — once it returns, mpv will never invoke the dead trampoline again. Fixes the **deterministic crash path** that all reported users are hitting.

### Layer 2: Debug-mode isolate fallback (`initializer.dart`)

In debug mode (where hot restart exists), use the isolate-based event loop instead of `NativeCallable`. The isolate approach polls with `mpv_wait_event` and **never registers a wakeup callback**, so there is nothing to become stale across restarts. Spawned isolates survive hot restart; messages to the dead main isolate's `SendPort` silently fail.

- `NativeCallable` is still used in **release/profile mode** (maximum performance, no hot restart)
- `NativeCallable` is still used in **unit test isolates** (tests the production code path)
- ~7-8 max concurrent players in debug mode due to isolate thread pool (acceptable for development)

### Layer 3: Defensive null-check in callback (`initializer_native_callable.dart`)

Early return in `_callback()` when context has been disposed, with double-check inside the synchronized lock.

## Why Previous Fix (PR #1350) Was Reverted

PR #1350 added Dart-level null-checks and shutdown synchronization but **kept NativeCallable in debug mode**. This can never fully work because the crash happens at the **native trampoline level** (freed memory), before any Dart code executes. No amount of Dart-level guards prevents the VM from detecting the UAF.

## Why This Fix Works

- **Layer 2 eliminates the root cause**: No `NativeCallable` in debug mode = no dead trampoline to invoke
- **Layer 1 fixes cleanup**: Even if NativeCallable was used (release mode, tests), stale handles won't crash during NativeReferenceHolder cleanup
- **Layer 3 adds safety**: Prevents event processing on disposed contexts during explicit `dispose()` races

## Test Plan

- [ ] Hot restart on Android (physical device) with Flutter 3.38+
- [ ] Hot restart on iOS simulator with Flutter 3.38+
- [ ] Hot restart on macOS with Flutter 3.38+
- [ ] Hot restart on Windows with Flutter 3.38+
- [ ] Hot restart on Linux with Flutter 3.38+
- [ ] Verify video playback works normally after hot restart
- [ ] Verify multiple sequential hot restarts don't accumulate player instances
- [ ] Run existing `dart test` suite (initializer_test.dart)
- [ ] Verify release mode still uses NativeCallable (performance)

Fixes #1340
Fixes #1314
Ref: https://github.com/flutter/flutter/issues/178548


realmeylisdev:
It's so annoying our whole team, please let's fix this ASAP!

Predidit:
This solution looks like a combination of my PR #1352 and jmgeffroy’s PR ~~#1386~~ #1416. Why is this mix necessary? These were originally two independent solutions.

birros:
Hello everyone, let's try to unblock this issue.

The patch looks clean to me and properly fixes the crash on hot restart.

If no one from the following list raises an objection within the next few days, I will merge this PR:

- @realmeylisdev
- @Predidit
- @alexmercerind (Pinging you, but feel free to ignore if you'd like 😉)
- @Carapacik

Merging this PR will also close the following PRs:

- https://github.com/media-kit/media-kit/pull/1352
- ~~https://github.com/media-kit/media-kit/pull/1386~~ -> https://github.com/media-kit/media-kit/pull/1416

Thank you all.

realmeylisdev:
> Hello everyone, let's try to unblock this issue.
> 
> The patch looks clean to me and properly fixes the crash on hot restart.
> 
> If no one from the following list raises an objection within the next few days, I will merge this PR:
> 
> * @realmeylisdev
> * @Predidit
> * @alexmercerind (Pinging you, but feel free to ignore if you'd like 😉)
> * @Carapacik
> 
> Merging this PR will also close the following PRs:
> 
> * [fix: fix flutter338 hot restart crash #1352](https://github.com/media-kit/media-kit/pull/1352)
> * ~~https://github.com/media-kit/media-kit/pull/1386~~ -> https://github.com/media-kit/media-kit/pull/1416
> 
> Thank you all.

Hey @birros would you mind resolving CI errors, thank you so much for merge consideration!

birros:
> Hey @birros would you mind resolving CI errors

@realmeylisdev Done (thx @Carapacik)

- https://github.com/media-kit/media-kit/pull/1414

Predidit:
@birros 

Thanks for pushing this forward.

After thinking about it a bit more, my preference would still be to merge ~~#1386~~ #1416 first. It is a smaller and more targeted fix for the deterministic crash path: clearing the mpv wakeup callback before sending `quit`.

My concern with this PR is that it combines two originally independent approaches: the debug-mode isolate fallback from #1352 and the native cleanup fix from ~~#1386~~ #1416. That makes the change larger and a bit harder to reason about.

I don’t want to block progress here, but I would feel more comfortable landing ~~#1386~~ #1416 as the minimal fix first, and then discussing whether the debug-mode isolate fallback is still needed as a follow-up.

birros:
Now that https://github.com/media-kit/media-kit/pull/1416 has been merged and properly fixes hot restart, are https://github.com/media-kit/media-kit/pull/1352 and this PR still relevant?

In short, what do https://github.com/media-kit/media-kit/pull/1352 and this PR still add now that hot restart is fixed in the main branch?

Predidit:
@birros Thanks for following up.

Now that #1416 has been merged and hot restart is properly fixed on `main`, I agree that #1352 is no longer necessary.

To clarify my earlier position: #1352 was mainly a workaround I proposed when I could not fully address the root cause yet. It avoided the problematic `NativeCallable` path in debug mode, but after #1416 fixed the actual cleanup issue by clearing the mpv wakeup callback before `quit`, I don’t think #1352 adds enough value to justify keeping that extra complexity.

So from my side, I’m fine with closing #1352. I also don’t think this PR needs to keep the debug-mode isolate fallback unless someone sees another concrete issue that #1416 does not cover.

Thanks again for helping move this forward.

Predidit:
@birros

Please wait a bit before closing this completely. In my testing, I still see a crash during hot restart, although the more important hot reload path no longer crashes.

I’d like to hear your opinion on whether this is enough for now. Hot restart is used much less often, and while a crash there can still confuse developers, it does not block normal development as badly as the previous hot reload crash.

Carapacik:
Hot restart still crashes for me with only the changes from #1416.

I tested this on Windows, macOS, iOS, and Linux. In the current state without the changes from this PR, hot restart still triggers a crash on all of these platforms in my setup.

If I also apply the changes from this PR, the crash is gone. So I think this PR, or at least an additional follow-up fix based on it, is still relevant.

Predidit:
I think there are really only two coherent options here:

1. Revert/replace #1416 and merge #1352 instead.
   That would be the workaround-based approach. It avoids the NativeCallable path in debug mode and should cover both hot-reload and hot-restart. But to be honest, even as the author of #1352, I don’t particularly like that solution. It is a workaround with clear trade-offs, not a clean fix.

2. Keep the current state after #1416.
   This is the cleaner option. #1416 fixes hot-reload properly, which is the main path developers use in day-to-day development. Hot-restart still crashes, but that path is used much less often, and fixing it cleanly may require support from Flutter that we currently do not have.

**What I strongly object to is merging both approaches together through #1392.**

**#1352 and #1416 are two different strategies. Combining them does not make the result stronger. Once the #1352 fallback is active, the relevant NativeCallable path fixed by #1416 is bypassed in debug mode, so the #1416 part becomes mostly meaningless for the main scenario affected by this issue.**

In other words, #1392 feels like taking the already-not-great workaround from #1352 and then adding a few lines from #1416 on top, even though those lines no longer matter for the debug-mode code path. That makes the final behavior harder to reason about, not better.

So my position is:

- If we want to cover both hot-reload and hot-restart, merge #1352 as the workaround and accept its trade-offs.
- If we prefer the cleaner fix, keep #1416, accept that hot-restart remains unsupported/crashy for now, and close #1352/#1392.
- But we should not merge #1392, because it forcibly combines two different solutions into one patch.

I’d like to hear your opinion on which direction you prefer.

Carapacik:
@birros Any updates?

birros:
> @birros Any updates?

Sorry for the delay :)

---

> **#1352 and #1416 are two different strategies. Combining them does not make the result stronger. Once the #1352 fallback is active, the relevant NativeCallable path fixed by #1416 is bypassed in debug mode, so the #1416 part becomes mostly meaningless for the main scenario affected by this issue.**
> …
> I’d like to hear your opinion on which direction you prefer.

@Predidit You are absolutely right, #1416 is the more solid fix but unfortunately it is not sufficient. I don't think the coexistence of #1416 and #1352 is a real problem, and even if it were, nothing critical since it only affects debug mode…

So, I will merge #1352 since you are the original author of the fixes copied into this PR. I'll close this PR now.

---

Last question to @Carapacik: the only difference between #1352 and this PR is the following diff, do you think it is necessary, is #1352 sufficient, or should I backport this part into #1352?

https://github.com/media-kit/media-kit/pull/1392/changes#diff-d1dc6e914df27cccc681de34d898cecf8f622d0078a02df545c4f02dfbe40c16

Carapacik:
I think #1352 is sufficient as the main fix.

Issue #1143, opened 2025-03-21
invert the width and height from videos


ksanchez9306:
Hi, i try to render videos from s3 using media_kit, the videos in s3 are generated using media converter, but in the device the video appear in miniature like in the image

![Image](https://github.com/user-attachments/assets/08cca7e9-15ee-49f3-b2f4-3edd0ae6c6ff)

the logs show this info

`videoParams: VideoParams(pixelformat: yuv420p, hwPixelformat: null, w: 1280, h: 720, dw: 1280, dh: 720, aspect: 1.7777777910232544, par: 1.0, colormatrix: bt.601, colorlevels: full, primaries: bt.601-625, gamma: bt.1886, sigPeak: 1.0, light: display, chromaLocation: mpeg2/4/h264, rotate: 0, stereoIn: mono, averageBpp: 12, alpha: null)

{rect: {top: 0, left: 0, width: 1280, height: 720}, wid: 18734, handle: -5476376647737350432, id: 7}`

but if i use ffprobe to get the information of the video storage in s3 return this 
command ` ffprobe -v error -select_streams v:0 -show_entries stream=width,height -of default=nw=1 aws_v3__c22f000f46894462901500c07c630e6a.temp.mp4_720p.tr.mp4`
width=720
height=1280

this is my logic for show the video widget
```final fit = aspectRatio.value != 1.0
                ? (aspectRatio.value > 1.0
                    ? BoxFit.fitWidth
                    : aspectRatio.value < 1.0
                        ? BoxFit.fitHeight
                        : BoxFit.cover)
                : BoxFit.cover;
            return Video(
              controller: _videoController,
              fit: fit,
              controls: (state) => widget.isCustomControls!
                  ? _buildPlayPauseOverlay(state)
                  : const SizedBox.shrink(),
            );```

only when change the fitWitdh by cover works in all the screen but the video looks like zoom like in the image

![Image](https://github.com/user-attachments/assets/8debf072-b9d4-4986-9a1f-27e3acb438b7)

but for example if choose the original video that is not transcode for different quality like 360,480 and 720p 
appear this 
[...]

PR #1405, opened 2026-04-13
fix: correct aspect ratio for rotated videos (#1143)

dyno-nexsoft:
## Overview
This document describes the fix for incorrect orientation and aspect ratio detection in `MediaKitVideoPlayer`, specifically affecting rotated videos such as portrait recordings from iPhones.

## The Problem
Portrait videos captured on many mobile devices (especially iPhones) are often stored as landscape frames (e.g., 1920x1080) with a metadata flag indicating a 90 or 270-degree rotation. 

Prior to this fix, the `VideoPlayerPlatform` implementation only utilized the raw display dimensions (`dw` and `dh`) from the `VideoParams` event, ignoring the `rotate` metadata. This resulted in:
1. The Flutter `video_player` plugin receiving incorrect dimensions (landscape instead of portrait).
2. The UI creating a landscape `AspectRatio` container.
3. The video appearing "shrunk" or "pillarboxed" incorrectly within the UI because the actual rendered content was vertical.

## The Solution
The implementation was updated to interpret the `rotate` property provided by `media_kit`. If the video is rotated by 90 or 270 degrees, the width and height are swapped before being emitted via the `VideoEvent.initialized` event.

### Implementation Detail
In `lib/src/media_kit_video_player.dart`:

```dart
// Check for rotation metadata
final rotate = event.rotate ?? 0;

// Swap dimensions for portrait-oriented videos stored as landscape
if (rotate == 90 || rotate == 270) {
  width = event.h;
  height = event.w;
} else {
  width = event.w;
  height = event.h;
}

// Notify the platform interface with corrected dimensions
if ((width ?? 0) > 0 && (height ?? 0) > 0) {
  notify();
}
```

This ensures that the `VideoPlayerController` in Flutter sees the correct effective size, leading to proper layout and scaling across all platforms.


dyno-nexsoft:
https://github.com/media-kit/media-kit/issues/1143

Issue #1155, opened 2025-04-09
MaterialSeekBarState.onPointerUp Null check operator used on a null value. Error thrown .


begovsky:
I get  following exception on some android devices when playing .m3u8 video:
```
Fatal Exception: io.flutter.plugins.firebase.crashlytics.FlutterError: Null check operator used on a null value. Error thrown .
#00 pc 0x381757 (MaterialSeekBarState.onPointerUp [framework.dart:959])
```

abdelaziz-mahdy:
are you using the latest release? i didnt find any null check in our code

begovsky:
@abdelaziz-mahdy I was not when I reported the crash but today I had an app update with the newest media-kit version. The old crash is still reproduced and not only that but the new one appeared. Please have a look. 

<img width="1263" alt="Image" src="https://github.com/user-attachments/assets/3ae2439f-98cc-45b9-9530-0287cdd8f7b4" />

abdelaziz-mahdy:
i think this is the only parts that has this problem 

<img width="599" alt="Image" src="https://github.com/user-attachments/assets/dcd51e0b-4d3e-4a3f-9fcd-17d2c0c4ac50" />

but it should not happen if the code is configured correctly, does it happen on the debug? since the assert should show if debugging

begovsky:
This is happening on production. 

<img width="1258" alt="Image" src="https://github.com/user-attachments/assets/e65e175d-644b-4a1c-8efe-88b2d0ad0ee1" />

This is my code:

```
class NativeVideoPlayer extends StatefulWidget {
  final String videoUrl;
  final String tvShowId;
  final Function onVideoLoaded;

  const NativeVideoPlayer({
    super.key,
    required this.videoUrl,
    required this.tvShowId,
    required this.onVideoLoaded,
  });

  @override
  State<NativeVideoPlayer> createState() => _NativeVideoPlayerState();
}

class _NativeVideoPlayerState extends State<NativeVideoPlayer> {
  late final player = Player();
  late final controller = VideoController(player);

  bool firstTimePlaying = false;

  StreamSubscription? _playingSubscription;

  void onFirstTimeVideoPlay(){
    widget.onVideoLoaded();

    trackWatchedTvShow(tvShowId: widget.tvShowId, videoLink: widget.videoUrl);
  }

  @override
  void initState() {
    super.initState();
    final autoPlay = context.read<EpisodePlayerCubit>().autoplay;

[...]

PR #1406, opened 2026-04-13
fix: prevent VideoStateInheritedWidget null check crash in seek bar and controls

kaankoken:
## Problem

`MaterialSeekBarState.onPointerMove` and several other callbacks call `controller(context)` → `VideoStateInheritedWidget.of(context)` → `result!`. When the widget tree rebuilds between a pointer event being queued and dispatched, `dependOnInheritedWidgetOfExactType<VideoStateInheritedWidget>()` can return `null` and the `!` throws `Null check operator used on a null value`.

The same anti-pattern exists in keyboard shortcut, scroll, tap, drag, long-press and seek-indicator callbacks across the `Material*` controls — any of them that runs after the inherited widget is momentarily gone from the ancestor chain will throw. Flutter recovers next frame, but it inflates the fatal Crashlytics rate.

## Fix

Cache the `VideoController` in a private `_controller` field, resolved in `didChangeDependencies` with an `identical()` check so a parent that swaps the controller correctly cancels the old subscriptions and re-attaches to the new player. Pointer / keyboard / scroll / drag / long-press / seek-indicator callbacks now use the cached field with `mounted` and null guards instead of looking up the inherited widget mid-event.

## Classes hardened

- `MaterialSeekBarState`
- `MaterialDesktopSeekBarState`
- `_MaterialVideoControlsState` (`_handleLongPress`, `_handleLongPressEnd`, `onHorizontalDragUpdate`, `onHorizontalDragEnd`, the `_BackwardSeekIndicator` / `_ForwardSeekIndicator` `onSubmitted` closures)
- `_MaterialDesktopVideoControlsState` (default `CallbackShortcuts` keyboard bindings, `Listener.onPointerSignal`, `GestureDetector.onTapDown` / `onPanUpdate`)
- `MaterialDesktopVolumeButtonState` (scroll, mute toggle, slider)

### DIFF
--- a/media_kit_video/lib/media_kit_video_controls/src/controls/material.dart
+++ b/media_kit_video/lib/media_kit_video_controls/src/controls/material.dart
@@ -7,6 +7,7 @@
 import 'dart:async';
 
 import 'package:flutter/material.dart';
+import 'package:media_kit/media_kit.dart' show Playlist;
 import 'package:media_kit_video/media_kit_video.dart';
 import 'package:media_kit_video/media_kit_video_controls/src/controls/extensions/duration.dart';
 import 'package:media_kit_video/media_kit_video_controls/src/controls/methods/video_state.dart';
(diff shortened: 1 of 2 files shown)


ymmtyuhei:
Hitting this in production on **Windows desktop** — the `MaterialDesktopSeekBarState` half of this PR.

```
TypeError: Null check operator used on a null value
culprit: State.context
FlutterErrorDetails.context: thrown while dispatching a pointer event

// dispatched from GestureBinding / RendererBinding
MaterialDesktopSeekBarState.build (package:media_kit_video/.../material_desktop.dart:1045)  // Listener closure
  onPointerMove                  (package:media_kit_video/.../material_desktop.dart:970)
    State.context                (package:flutter/src/widgets/framework.dart:959)  // _element! is null
```

- `media_kit_video` 2.0.1 (current pub latest), `media_kit` 1.2.6
- Flutter 3.44.8, release build, Windows Build 26200

`setState` is overridden to check `mounted`, but the `controller(context)` call right after it isn't — so the guard is bypassed. Same in `onPointerUp`.

Re [#1155](https://github.com/media-kit/media-kit/issues/1155) ("shouldn't happen if the app is configured correctly"): apps can't avoid it. `GestureBinding` caches the `HitTestResult` at pointer down and reuses it for every move/up until release, so the seek bar keeps receiving events after its `State` is disposed — and whether a drag is in progress is media_kit's internal state, invisible to the app. Any subtree removed while a button is held triggers it: a window resize crossing a breakpoint, a step advancing from device state. Release-only, since debug just asserts.

Happy to test a Windows build if that helps.

PR #1419, opened 2026-05-21
feat: add iosManageAudioSession option to PlayerConfiguration

### LINKED ISSUE #964: Multiple videos not playing correctly

### LINKED ISSUE #1205: on IOS mediakit completely breaks flutter webRTC call audio
After upgrading to media kit, our call audio stopped working completely in flutter webRTC. The problem is that mediakit uses the single AVAudioSession, and the player is conflicting with control of the AVSession in iOS. So even if I configure the audio session carefully stopping or starting another player disrupts control. 

The solution was to work around this directly with native code and use an alternate player in our case, there's one specific way.

### LINKED ISSUE #1227: Background audio playback on iOS not working.
Hi! First of all, thank you for the wonderful plugin. I implemented a mix of media_kit and audio_service working fine on Android. However, I was not able to make it work on iOS. I did configure everything audio_service requires for iOS to work, but still, no results.

I did look around and saw you were planning a plugin like media_kit_os_controls. I don't know if this actually happened. 
I also found that you mentioned https://github.com/media-kit/media-kit/issues/560#issuecomment-1785568310 that it works (only?) on Android. I also found this comment: https://github.com/media-kit/media-kit/issues/970#issuecomment-2857988972 regarding iOS.
 
Any ideas? Thanks again!

IgorKhramtsov:
## Problem
On iOS, mpv's ao_audiounit driver unconditionally configures and activates the process-wide `AVAudioSession`:
  - on init it calls `setCategory:` and `setActive:YES`
  - on uninit it calls `setActive:NO`

`AVAudioSession` is shared across the whole app, so this fights any other component that needs to manage it — e.g. `audio_session`, `just_audio`, other instance of `media_kit` player. The most visible symptom is that disposing a Player deactivates the session, which silences whatever else was playing in the host app.

## Change
Adds an `iosManageAudioSession` flag to `PlayerConfiguration` (default: `true`, behavior unchanged). When set to `false`, media_kit forwards `audiounit-skip-session-management=yes` to libmpv, which short-circuits the session calls inside `ao_audiounit` — leaving `setCategory`, `setActive:YES`, and `setActive:NO` under the embedder's control.

```dart
final player = Player(
  configuration: const PlayerConfiguration(
    iosManageAudioSession: false, // embedder owns AVAudioSession
  ),
);
```

The flag has no effect on platforms other than iOS.

## Demo
A standalone reproduction with fix demonstration is here: https://github.com/IgorKhramtsov/media_kit_ios_audio_session_example
It runs `just_audio` alongside `media_kit`. Disposing the `media_kit` Player silences the `just_audio` track when `iosManageAudioSession: true`, and leaves it playing when `false`.

## Related artifact work
The `audiounit-skip-session-management` option is not part of upstream mpv. It is added by a small patch applied during the `libmpv` build.

- https://github.com/media-kit/libmpv-darwin-build/pull/52 — PR adding `mpv-audiounit-skip-session-management.patch`

Fixes
  - Fixes #1205
  - Fixes #1227
  - Fixes #964

### DIFF
--- a/media_kit/lib/src/player/native/player/real.dart
+++ b/media_kit/lib/src/player/native/player/real.dart
@@ -2332,6 +2332,10 @@ class NativePlayer extends PlatformPlayer {
         // Set --vid=no by default to prevent redundant video decoding.
         // [VideoController] internally sets --vid=auto upon attachment to enable video rendering & decoding.
         if (!test) 'vid': 'no',
+        // Skip mpv's AVAudioSession management when the embedder owns the
+        // session. iOS-specific.
+        if (Platform.isIOS && !configuration.iosManageAudioSession)
+          'audiounit-skip-session-management': 'yes',
       };
 
       if (Platform.isAndroid &&
--- a/media_kit/lib/src/player/platform_player.dart
+++ b/media_kit/lib/src/player/platform_player.dart
@@ -514,6 +514,23 @@ class PlayerConfiguration {
   /// Learn more: https://ffmpeg.org/ffmpeg-protocols.html#Protocol-Options
   final List<String> protocolWhitelist;
 
+  /// Whether [Player] should let mpv configure and activate the global
+  /// `AVAudioSession` on iOS.
+  ///
+  /// When `false`, the AudioUnit audio output skips `setCategory`,
+  /// `setActive:YES` on init, and `setActive:NO` on dispose — leaving the
+  /// session under the embedder's control.
+  ///
+  /// `AVAudioSession` is process-wide on iOS, so when more than one
+  /// component manages audio (e.g. another player or a custom coordinator),
+  /// only one of them should own the session — otherwise their category and
+  /// activation calls will conflict and cause playback issues.
+  ///
+  /// No effect on platforms other than iOS.
+  ///
+  /// Default: `true`.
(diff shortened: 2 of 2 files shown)


alexmercerind (approved):

alexmercerind:
Thanks! I think it is good to have since it makes the library behave same on all platforms. Managing audio service is expected to be done externally.

IgorKhramtsov:
@alexmercerind hi!
Just a heads up, there is a build-side PR with patch that needs to be merged in order for this to work - https://github.com/media-kit/libmpv-darwin-build/pull/52

libmpv has no such capability out of the box for some reason

alexmercerind:
Thanks for informing. I would've missed this. I was also a bit confused since I was searching for similar option some time ago & didn't find any.

PR #1422, opened 2026-06-03
fix(windows): Fix Media.extras values always get deleted on windows

jpangestu:
Description
Fixes an issue on Windows where Media.extras metadata is lost whenever I do setShuffle.

Root Cause (According to Gemini)
During native playlist updates, the C++ bridge reconstructs Dart Media objects from raw paths. On Windows, these paths often contain backslash/forward slash mismatches or the extended path namespace prefix (e.g. //?/D:/Music/...). Because Media.normalizeURI doesn't standardize these Windows-specific path formats, the internal Media.cache lookup fails to resolve the original extras mapping.

Fix
Updated Media.normalizeURI under the URIType.file case for Windows to:

Strip Win32 namespace extended path prefixes (\\?\ or //?/).
Standardize all slashes to backslashes (\).
Convert the path to lowercase for case-insensitive lookup.

### DIFF
--- a/media_kit/lib/src/models/media/media_native.dart
+++ b/media_kit/lib/src/models/media/media_native.dart
@@ -168,7 +168,13 @@ class Media extends Playable {
     switch (parser.type) {
       case URIType.file:
         {
-          return parser.file!.path;
+          var path = parser.file!.path;
+          if (Platform.isWindows) {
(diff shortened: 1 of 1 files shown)


jpangestu:
Extras works fine on linux but not on windows. I tested this fix on my own music player app.
*I don't know if this helpful, this is my first ever pull request

